PR #2, opened 2020-06-09
Add icons to demo site

kjgalvan:
Adds simple favicon that shows the letter P (short for Prints of Poe)

Adds appropriately sized logos based on an svg I created to avoid licensing problems.


kjgalvan:
Closing because work was done on master

PR #6, opened 2020-06-11
Add carousel

kjgalvan:
Adds a slideshow carousel to the top of the home page with links to the stubbed out product page.

Adds responsive css styling so that carousel text is readable on smaller screens.

Slideshow images were edited with GIMP and Polarr editing software.

Image of site carousel


demo/src/screens/HomeScreen/HomeCarousel.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @return {!JSX} Slideshow component for showcasing
+ *     our latest deals and offerings.
+ */
+export function HomeCarousel() {
+  return (
+    <Carousel>
+      <Carousel.Item>
+        <img
```
zikas, line 16: In another PR I saw using Image from 'react-bootstrap/Image';
Ideally we should stay consistent all over the codebase.

btw, how className would work for the regular "img" tag? (might be my gap in understanding react)

demo/src/screens/HomeScreen/HomeCarousel.js
```diff
@@ -0,0 +1,50 @@
...
+          src={slideOne}
+          alt="First slide"
+        />
+        <Carousel.Caption>
+          <h3>Stunning High Quality Prints Available</h3>
+          <p>View our <a href="/product/123">latest piece</a>.</p>
+        </Carousel.Caption>
+      </Carousel.Item>
+      <Carousel.Item>
+        <img
```
zikas, line 27: ditto

demo/src/screens/HomeScreen/HomeCarousel.js
```diff
@@ -0,0 +1,50 @@
...
+          src={slideTwo}
+          alt="Third slide"
+        />
+        <Carousel.Caption>
+          <h3>SALE!</h3>
+          <p>Get this classic 50% off. <a href="/product/123">View Deal</a></p>
+        </Carousel.Caption>
+      </Carousel.Item>
+      <Carousel.Item>
+        <img
```
zikas, line 38: ditto

demo/src/screens/HomeScreen/HomeCarousel.js
kjgalvan: I definitely agree we should stay consistent. I'll go ahead and change to Image :)

The className thing is a React weirdness thing - JSX mimics HTML in a lot of ways but is ultimately a flavor of Javascript where class is a reserved word. [Here's some reading if you're curious](https://reactjs.org/docs/dom-elements.html)!

zikas: Great, thanks for the link!

PR #7, opened 2020-06-11
Add all 30 product images and thumbnails

kjgalvan:
Adds 30 edited "high quality" images for our print products. These are to be displayed on the product page only due to their large size.

Adds 30 1:1 ratio thumbnails for all products. These are to be displayed on the main page in product listings and the shopping cart page for item previews.


calhounsm:
One thing to note. GitHub's guidance is no files greater than 100MB and try to keep the total repo size under 1GB.  https://github.community/t/working-with-large-files-and-repositories/10203

This commit eats up a little under half of that total size guidance.

Approved since I don't expect there will be too many more large files to commit in this repo.

PR #9, opened 2020-06-12
Adding myself to the AUTHORS file.

durbanie:


AUTHORS
```diff
@@ -0,0 +1,6 @@
+# This is a list of contributors
+
+# Names should be added to this file like so:
+# Name or Organization <email address>
+
+* Dustin Urbaniec <dustin.urbaniec@gmail.com>
```
zikas, line 6: Should everyone add themself individually? or we can just put all contributors here at once?

durbanie: I was thinking everyone would add themselves individually, so they can choose the email to use. Plus I was hoping to test the review list, but looks like it isn't working.

Hold off on this for now... I'm going to try a different approach.

PR #10, opened 2020-06-12
Create Minicart component for checkout screen

Alex7Li:
In this pull request I add and style a smaller version of the cart component. The smaller cart will provide a way for users to see the contents of their cart without having to go back to the cart screen. It has been added as the second half of the checkout screen (along with a form asking for user data which is still in progress). Since the minicart and the actual cart are closely related, I ended up styling the cart component as well. 

The images below show what the checkout and cart pages look like on a large and medium sized screen.
Cart Large
Screenshot 2020-06-12 at 4 14 58 PM
Cart Medium
Screenshot 2020-06-12 at 4 15 20 PM
Checkout Large
Screenshot 2020-06-12 at 4 11 20 PM
Checkout Medium
Screenshot 2020-06-12 at 4 11 11 PM


Alex7Li:
Requesting a review from @googleinterns/measurement-library-owners

demo/src/components/Cart/Cart.js
```diff
@@ -5,13 +5,28 @@ import './Cart.css';
 import {setQuantity} from '../../store/StoreHelpers.js';
 import PropTypes from 'prop-types';
 
+/**
+ * Computes the price of purchasing the given quantity
+ * of all items in the cart.
+ * @param {!Object<string,{cost:number, quantity:number}>} items The items
```
calhounsm, line 11: It seems like this is essentially the same value type as used in CartBase's items parameter. I'd recommend pulling it out to a typedef, rather than just declaring it one-off for each method.

https://google.github.io/styleguide/jsguide.html#jsdoc-enum-and-typedef-comments

demo/src/components/MiniCart/MiniCart.js
```diff
@@ -0,0 +1,60 @@
...
+import React from 'react';
+import {computePrice} from '../Cart/Cart.js';
+import '../Cart/Cart.css';
+
+/**
+ * Creates a component describing a shopping cart.
+ * @param {!Object<string,
+ *      {name:string, item:!Object, quantity:number, description:string,
+ *      inCart:boolean, cost:number}>} items The items stored in the site state.
+ * @param {function(string, number)} setQuantity A function to modify
```
calhounsm, line 13: It looks like this method does not have a setQuantity param.

demo/src/components/MiniCart/MiniCart.js
```diff
@@ -0,0 +1,60 @@
...
+import {connect} from 'react-redux';
+import PropTypes from 'prop-types';
+import React from 'react';
+import {computePrice} from '../Cart/Cart.js';
+import '../Cart/Cart.css';
+
+/**
+ * Creates a component describing a shopping cart.
+ * @param {!Object<string,
+ *      {name:string, item:!Object, quantity:number, description:string,
```
calhounsm, line 11: Likewise, if this item type is used across several modules, consider which it would be best to define it in, or even create a new "types" module for holding types used around the site.

Alex7Li: I've added a typedef in store/store.js since that's where the object is defined. No module since I don't think we will have any other big types.

demo/src/components/MiniCart/MiniCart.js
Alex7Li: Removed it.

demo/src/components/Cart/Cart.js
Alex7Li: Refactored

demo/src/components/Cart/Cart.js
```diff
@@ -5,13 +5,28 @@ import './Cart.css';
...
+ * to purchase and their quantity.
+ * @return {number} The total price.
+ */
+export const computePrice = (items) => {
+  let totalPrice = 0;
+  for (const [, item] of Object.entries(items)) {
```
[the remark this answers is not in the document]
Alex7Li: Refactored

PR #16, opened 2020-06-15
Add Code Modal component to highlight and display analytics code

Alex7Li:
This commit adds a modal component to view highlighted java script code corresponding to google analytics commands. This is needed so that the user can see what command a button click on the demo page corresponds to.
Screenshot 2020-06-15 at 11 02 56 AM
Screenshot 2020-06-15 at 11 02 47 AM

(Same as #15 but replaced to test the auto assign reviewer functionality, which works)


demo/src/vendor/prism.css
```diff
@@ -0,0 +1,152 @@
...
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * @FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
```
durbanie, line 33: This could be a problem if this ever becomes an officially Google supported project. While the Apache 2 and MIT LICENSEs are compatible, this will cause difficulties for Google's Open Source licensing administrators and legal teams.

Is there another way you can do what you need here without directly copying this code into the project? Is there a way you can depend on this CSS instead?

Alex7Li: I found an alternative way to import the css in the documentation.

demo/src/components/CodeModal/CodeModal.js
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * A popup to display the google analytics code that fires on
+ * a button click.
+ * @param {string} popupId The ID of this popup.
+ * @param {string} ourCode The code for this open source library.
+ * @param {string} gtagCode The code for gtag.
+ * @return {!JSX} The component.
+ */
+export function CodeModal({popupId, ourCode, gtagCode}) {
```
durbanie, line 22: nit: I'd change this to something like "measurementLibraryCode"

Alex7Li: I was planning to go back and change it once we've decided on an official name, I'll bring this up  at the standup today

Alex7Li: Fixed

PR #17, opened 2020-06-15
structure for product screen

pedrosilva0111:
Add presentational structure for the product page.

image

### DIFF
--- a/demo/src/screens/ProductScreen/ProductScreen.js
+++ b/demo/src/screens/ProductScreen/ProductScreen.js
@@ -1,4 +1,6 @@
 import React from 'react';
+import {Container, Row, Col, Image, Button} from 'react-bootstrap';
+import blackDemo from '../../images/black.png';
 
 /**
  * @return {!JSX} Page component for where a user can view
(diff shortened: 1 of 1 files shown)


demo/src/screens/ProductScreen/ProductScreen.js
```diff
@@ -1,13 +1,49 @@
...
  *     details about a single product.
  */
 export function ProductScreen() {
   return (
-    <div>
-      <h1>{`I'm the Product Page!`}</h1>
-    </div>
+    <Container>
+      <Row>
+        <Col/>
```
calhounsm, line 13: What is the reason for the empty first and last columns?  (Or am I misinterpreting this?)

pedrosilva0111: They are meant to split the row into three equally sized columns so that the content is centered on the screen.

calhounsm: My impression was that CSS is preferred for layout considerations like that over pure HTML implementations like tables, however I'm not an expert on web FE by any means.

I did do a quick look around and found some suggestions for how this could be achieved using some CSS:
Flexbox, for distributing space along an axis - https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Basic_Concepts_of_Flexbox
CSS grids - for more complex grid-based layouts - https://developers.google.com/web/updates/2017/01/css-grid?hl=en

I'm not gonna push much on this though, so if you'd prefer using a table that's fine, just reply here and I'll merge the PR.

pedrosilva0111: I also am fairly new to this framework. I did some reading on the conainer, row, and column components from react bootstrap and it seems like the framework uses CSS flexbox under the hood. https://react-bootstrap.github.io/layout/grid/ I will change it to CSS if that is better though!

calhounsm: Ah, I think that's good then, no need to reinvent the wheel if React already does the right thing for us.  Thanks for looking into it, and bearing with me on basic questions like this :).

PR #18, opened 2020-06-16
Add gtag

kjgalvan:
Adds gtag with helper functions to send events.

Sends page_view events automatically based on our router. Note that we disabled the automatic tracking of page_views based on our browser's history as manually sending the page_view is more consistent.


demo/src/lib/gtag.js
```diff
@@ -0,0 +1,24 @@
...
+ * @param {string} url
+ */
+export const pageview = (url) => {
+  window.gtag('config', GA_TRACKING_ID, {
+    page_path: url,
+  });
+};
+
+/**
+ * Sends an event to Google Analytics.
```
zikas, line 16: A small note.
It will send an event to all of the targets previously configured using "config" command (unless parameters specify the target explicitly using "send_to" property).
For example if there's a gtag('config', 'AW-123') called before, it will also send data to Google Ads. If no "config" has been called then nothing will be sent.
So the statement might not always be true and depends on previous invocations as well as parameters value.

You should consider rephrasing or adding `parameters['send_to'] = GA_TRACKING_ID` in the method to guarantee that data will be sent to your GA property.

demo/src/lib/gtag.js
```diff
@@ -0,0 +1,24 @@
...
+ * View the [Developer Guide](https://developers.google.com/analytics/devguides/collection/app-web/events)
+ * to review recommended events and their associated parameters.
+ * @param {action} event
```
zikas, line 19: Action is not a valid type, should be string

demo/src/lib/gtag.js
```diff
@@ -0,0 +1,24 @@
...
+ * @param {action} parameters
```
zikas, line 20: Action is not a valid type, should be !Object<string, *>

demo/src/lib/gtag.js
```diff
@@ -0,0 +1,24 @@
+
+// eslint-disable-next-line no-undef
+const GA_TRACKING_ID = process.env.REACT_APP_GA_TRACKING_ID;
+
+/**
+ * Sends a page_view event to Google Analytics.
+ * @param {string} url
```
zikas, line 7: Please add parameter description

Per https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments
Method, parameter, and return descriptions (but not types) may be omitted if they are obvious from the rest of the method’s JSDoc or from its signature.

In this case it's not obvious that provided url will be used as a page path. Even more confusing that it's called "url" but the value should be the path part of the url (full page url is not acceptable value), consider renaming to pagePath (might work without a description)

demo/src/lib/gtag.js
```diff
@@ -0,0 +1,24 @@
+
+// eslint-disable-next-line no-undef
```
zikas, line 2: It's not common to suppress lint warnings, are you sure it cannot be avoided?

demo/src/lib/gtag.js
```diff
@@ -0,0 +1,24 @@
...
+};
+
+/**
+ * Sends an event to Google Analytics.
+ * View the [Developer Guide](https://developers.google.com/analytics/devguides/collection/app-web/events)
+ * to review recommended events and their associated parameters.
+ * @param {action} event
+ * @param {action} parameters
+ */
+export const event = ({event, parameters}) => {
```
zikas, line 22: I'm confused, the jsdoc implies two arguments, but the method definition specifies a single object parameter.. where's the truth?

demo/src/lib/gtag.js
kjgalvan: Thank you, I didn't realize that is how it worked to determine where to send events to. I added the explicit parameter.

demo/src/lib/gtag.js
kjgalvan: I did some more research into this - I usually have an empty [dependency array](https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects) to make it very explicit that the code runs exactly once but it seems this has fallen out of favor.

The history object should be safe to pass into the array as it should never change ensuring the code runs exactly once. Some testing using the Debug Mode seems to confirm this.

kjgalvan: Whoops I originally though this referring to a different part of the PR. Ignore any previous notifications on this. I looked into a bit more and adjusted the eslint config to allow access to process. Thanks!

kjgalvan:
Apologies for adding a bit more code - I had forgot I had agreed to return a code snippet for our informational modals from the event function. It's an easy way we can capture the events being sent and display their gtag counterpart.

demo/src/lib/gtag.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Sends an event to Google Analytics.
+ * Can be spoofed to return mock code snippet of event instead.
+ * View the [Developer Guide](https://developers.google.com/analytics/devguides/collection/app-web/events)
+ * to review recommended events and their associated parameters.
+ * @param {string} event
+ * @param {!Object<string, *>} parameters
+ * @param {boolean=} spoof If true, returns mock code instead of sending event.
+ * @return {?string} Code snippet of the event being sent.
```
zikas, line 21: Not particularly wrong but the provide boolean essentially splits this method into two. It does look like a violation of single responsibility principle and it's a bit hard to reason about what this the main purpose of this method and at what circumstances the return is expected.

I'm not sure if supporting such spoofing is even necessary but in case it is
consider having two separate methods or always returning the code snippet (so caller doesn't have to null-check the return value).

kjgalvan: Yeah this was debated a bit amongst ourselves. We don't want it to be returned every time because we will need the code snippet before firing off the event in a lot of cases, but we also were unsure of splitting the function up because there are [several functions built on top of this one which we would also need to split up](https://github.com/googleinterns/measurement-library/pull/25/files#diff-2bfb813f8773f5d91020c916ca5ae49d).

I do agree it violates the SRP and can indeed split this and the other functions up if you recommend it, though!

kjgalvan: To add a bit more context: We are trying to build code snippets based on the state of our site in the same way we build events based off of the site state.

kjgalvan: As an update: I went ahead and separated out the two functions since it is indeed two different things we are trying to do :)

demo/src/Router.js
```diff
@@ -1,20 +1,34 @@
...
 import {ProductScreen} from './screens/ProductScreen/ProductScreen.js';
 import {ThankScreen} from './screens/ThankScreen/ThankScreen.js';
+import {pageview} from './lib/gtag.js';
 
 /**
- * @return {!JSX} Router component that switches which page is
- *     displayed based on the URL. Defaults to home page.
+ * Creates component that switches which page is displayed based on the URL.
+ * Defaults to home page. Sends page_view event to Google Analytics.
+ * @return {!JSX}
```
zikas, line 19: Should you have @param?

PR #21, opened 2020-06-16
Create about screen

Alex7Li:
Users in the site may be a bit confused as to what it is for. The about screen clarifies this and provides some resources. Mostly copied from https://ga-dev-tools.appspot.com/enhanced-ecommerce/. 

Screenshot 2020-06-16 at 5 56 47 PM

(Also changed a line of code in the product screen that referenced to a non-existing file)

### DIFF
--- a/demo/src/screens/AboutScreen/AboutScreen.js
+++ b/demo/src/screens/AboutScreen/AboutScreen.js
@@ -13,7 +13,7 @@ const /** JSX */ DESCRIPTION = <p>{'This page is a sample ecommerce ' +
   'a button press, click the '}<GoQuestion size={16}/>
 {' icon next to the button. This will bring up code samples for both the '}
 <a href={'https://developers.google.com/gtagjs'}>{'gtag.js '}</a>
-{'library and the TODO:NAME_THIS library. ' +
+{'library and the measure library. ' +
   'In addition to showing sample code, this site sends real data to Google ' +
   'Analytics. To inspect the hits, just open up the developer ' +
   'tools or use the '}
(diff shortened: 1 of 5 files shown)


demo/src/screens/AboutScreen/AboutScreen.js
```diff
@@ -1,12 +1,41 @@
 import React from 'react';
+import {GoQuestion} from 'react-icons/go';
+import './AboutScreen.css';
+
+const /** JSX */ DESCRIPTION = <p>{'This page is a sample ecommerce ' +
+'application with a complete implementation of '}
```
zikas, line 6: This is probably me not understanding JSX but why <p>{'Text'} instead of <p>Text
https://reactjs.org/docs/introducing-jsx.html#embedding-expressions-in-jsx

for multiline html it seems that wrapping with () should be sufficient 
https://reactjs.org/docs/introducing-jsx.html#jsx-represents-objects

demo/src/screens/AboutScreen/AboutScreen.js
```diff
@@ -1,12 +1,41 @@
 import React from 'react';
+import {GoQuestion} from 'react-icons/go';
+import './AboutScreen.css';
+
+const /** JSX */ DESCRIPTION = <p>{'This page is a sample ecommerce ' +
```
zikas, line 5: !JSX

demo/src/screens/AboutScreen/AboutScreen.js
Alex7Li: Some places consider it a best practice to use {'text'} all the time since it can handle special characters and whitespace a bit better. I don't think that's a  very common practice (I didn't see it in the airbnb style guide for example), and it really clutters up this screen without providing much benefit, so I took it out for the most part. I left it inside of the <a> tags due to whitespace issues.

demo/src/screens/AboutScreen/AboutScreen.js
Alex7Li: fixed

PR #22, opened 2020-06-17
Add utility functions for building gtag events

kjgalvan:
Adds a utility function for building an array of these parameters from the items currently in the cart. Will be used to create the items parameter array in many gtag events.

Adds a utility function for returning some useful item parameters from an item in the redux store. If a user wants to add more parameters than these to their items array, they can still use this function to help build their items parameter array.

### DIFF
--- a/demo/src/utils.js
+++ b/demo/src/utils.js
@@ -61,7 +61,7 @@ export function getItemParameters(itemId) {
   */
 export function getItemsArrayFromCart() {
   const /** Array<!ItemParameters> */ itemsArray = [];
-  const items = store.getState().items;
+  const /** {ItemStore} */ items = store.getState().items;
   for (const [itemID, item] of Object.entries(items)) {
     if (item.inCart) {
       itemsArray.push({
--- a/demo/src/utils.js
+++ b/demo/src/utils.js
@@ -30,10 +30,13 @@ export function deepCopy(data) {
   */
 export function getItemParameters(itemId) {
   const /** {ItemStore} */ items = store.getState().items;
+  const /** {number} */ quantity = items[itemId].quantity;
   return {
     item_id: itemId,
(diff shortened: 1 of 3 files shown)


demo/src/utils.js
```diff
@@ -12,3 +13,61 @@
...
+  */
+export function getItemParameters(itemId) {
+  const /** {ItemStore} */ items = store.getState().items;
+  const /** {number} */ quantity = items[itemId].quantity;
+  return {
+    item_id: itemId,
+    item_name: items[itemId].name,
+    price: items[itemId].cost,
+    // if quantity is positive, include it in the item parameter
+    ...(quantity > 0 && {quantity}),
```
durbanie, line 53: Nifty!

demo/src/utils.js
```diff
@@ -12,3 +13,47 @@
...
+/**
+ * Type definition for our application's standard item parameters.
+ * These are to be sent in the items array found in many gtag events.
+ * Acceptable parameters can be [found here](https://developers.google.com/gtagjs/reference/aw-events)
+ * @typedef {{item_id: string, item_name: string,
+  *     price: string}} ItemParameters
+  */
+
+/**
+  * Selects an item from the store via ID and returns an object containing
```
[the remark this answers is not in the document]
kjgalvan: Great catch! Pretty sure this was a result of some weird auto formatting when copy+pasting. I've updated the PR to fix the indentations.

demo/src/utils.js
kjgalvan: Thanks! ES6 has some fun stuff. Of course it isn't as readable hence the comment but it gets the job done :)

durbanie: Yep, as long as you comment, I'm good with this ES6 "magic".

PR #19, opened 2020-06-16
Functional product screen

pedrosilva0111:
Pulls data from global store and displays correct info based on product ID. Add to cart and buy now functionality also added.

PR #23, opened 2020-06-17
Minor bug fixes

Alex7Li:
Some minor bug fixes.

1. When users hit enter on the cart page, the page would refresh, which is not desirable. Now it does nothing.

2. Some sample data violated the assumption that inCart=false => quantity=0. I set all the inCart flags to false and quantity to 0 since the site is almost ready. Because of this, this PR shouldn't be merged until #19 is merged, as otherwise it may become hard to test any functionality requiring the cart to be nonempty.

3. Remove a line that threw an error unnecessarily

4. When thumbnails were made, the cart screen and jsdoc wasn't updated to match it.


demo/src/components/Cart/Cart.js
```diff
@@ -20,6 +20,7 @@ export const computePrice = (items) => {
   return totalPrice;
 };
 
+
```
calhounsm, line 23: Nit: Unnecessary line break.  Only a single break is required between any two statements:
https://google.github.io/styleguide/jsguide.html#jsdoc-whitespace

demo/src/store/store.js
```diff
@@ -6,12 +6,12 @@ import {initialState} from './initialState.js';
  * Modifies the state of the store according to a description of the
  * changes to make given in the action parameter.
  * @param {{items:ItemStore}} state The state to take in initially.
- * @param {{type:string, id:string, amount:?number}} action a description of
+ * @param {{type:string, id:string, amount:?number}} action A description of
  * the action to take.
```
calhounsm, line 10: Nit: Add 4 spaces of indentation here.
https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments

demo/src/store/store.js
```diff
@@ -22,17 +22,17 @@ function reducer(state = {}, action) {
     case 'DECREMENT':
       item.quantity--;
       if (item.quantity < 0) {
-        console.error('Setting quantity to a negative amount!');
-        item.quantity++;
+        // Set quantity to 0 as negatives make no sense.
```
calhounsm, line 25: A console.warn might still be appropriate here, since this indicates a possible bug elsewhere in the application.  console.error was probably not necessary, though.

Alex7Li: Oh, I didn't know about console.warn, I'll use that in the future. I'm don't think that we should fire off any warnings here because users can call this code a lot - in the cart screen they can decrement the amount of things in their cart or make it a negative number.

demo/src/components/Cart/Cart.js
Alex7Li: Removed

demo/src/store/store.js
Alex7Li: Added

demo/src/store/store.js
calhounsm: Ah, fair enough.  If this is validating user input and preventing them from going negative on quantity then agreed, no warning needed.

PR #24, opened 2020-06-17
Split checkout screen to add billing information component

Alex7Li:
This PR adds forms to ask the user for their billing information, which appear on the checkout page after user information is submitted. This will allow for us to send more analytics events.

Screenshot 2020-06-17 at 2 51 56 PM
Screenshot 2020-06-17 at 2 51 41 PM

### DIFF
--- a/demo/src/components/BillingInfoForm/BillingInfoForm.js
+++ b/demo/src/components/BillingInfoForm/BillingInfoForm.js
@@ -1,0 +1,52 @@
+import React from 'react';
+import {Form} from 'react-bootstrap';
+import PropTypes from 'prop-types';
+import {makeFormGroup} from '../UserInfoForm/UserInfoForm.js';
+
+const /** !JSX */ cardType = makeFormGroup(/** label= */ 'Credit Card Type',
+    /** control= */ <Form.Control as="select" custom>
+      <option>Mastercard</option>
+      <option>Visa</option>
+      <option>American Express</option>
+      <option>Discover</option>
+    </Form.Control>,
+);
+
+const /** !JSX */ cardNumber = makeFormGroup(/** label= */ 'Credit Card Type',
+    /** control= */ <Form.Control type='number' required
+      defaultValue='4111111111111111'
+      placeholder='Credit Card Number (required)'
+    />,
+);
+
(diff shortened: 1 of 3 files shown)


demo/src/components/BillingInfoForm/BillingInfoForm.js
```diff
@@ -0,0 +1,52 @@
...
+        defaultValue='19123'
+        placeholder='Zip (optional)'
+      />, /** useWholeRow= */ false)}
+</Form.Row>;
+
+/**
+ * Creates a page component with a form to get billing information.
+ * @return {!JSX} The component.
+ */
+export function BillingInfoForm({formId}) {
```
zikas, line 40: I see that sometimes methods accept parameter object, sometimes the list of individual parameters, how do you decide where to use each one?

demo/src/components/UserInfoForm/UserInfoForm.js
```diff
@@ -10,7 +10,7 @@ import PropTypes from 'prop-types';
  *      column by itself.
  * @returns {!JSX} A form group with the given label and control.
  */
-const makeFormGroup = (label, control, useWholeRow=true) => {
+export const makeFormGroup = (label, control, useWholeRow=true) => {
```
zikas, line 13: Should be `useWholeRow = true`
https://google.github.io/styleguide/jsguide.html#features-functions-parameter-return-types

demo/src/components/BillingInfoForm/BillingInfoForm.js
Alex7Li: The only times that I've done it are when accepting arguments from JSX (like here) or the redux store. Generally speaking, I think it's good to use a parameter object so that every field of the object has a specific individual purpose. With JSX, I could accept a props parameter, but then every time I used said parameter I would be writing props.fieldname, so it's more clear to expand it at first. With the redux store, they always pass in an object with at least 2 fields, representing data and functions to change said data, since these feel like they have 2 different purposes, I split them up.

demo/src/components/UserInfoForm/UserInfoForm.js
Alex7Li: Added spaces

PR #25, opened 2020-06-17
Add ecommerce event helper functions

kjgalvan:
Adds basic ecommerce events that rely on PR #18 and PR #22 to work.

These will be added to the appropriate places in our application. More customized events like view_promotion will be created separately and in the relevant React component.


demo/src/lib/gtagEvents.js
```diff
@@ -0,0 +1,132 @@
...
+ * to review the event and its accepted parameters.
+ * @param {string} itemId
+ */
+export function sendSelectItemEvent(itemId) {
+  event('select_item', {
+    items: [{
+      ...getItemParameters(itemId),
+    }],
+  });
+}
```
durbanie, line 17: nit: This pattern is repeated multiple times. Maybe DRY it up by creating a module-level helper function?

demo/src/lib/gtagEvents.js
```diff
@@ -0,0 +1,132 @@
...
+ * View the [Ecommerce Guide](https://developers.google.com/analytics/devguides/collection/app-web/ecommerce)
+ * to review the event and its accepted parameters.
+ */
+export function sendViewCartEvent() {
+  event('view_cart', {
+    value: computePriceOfItemsInCart(),
+    currency: 'USD',
+    items: getItemsArrayFromCart(),
+  });
+}
```
durbanie, line 72: nit: This pattern is also repeated (and extended below). Any chance you can think of a way to dry this up?

Not necessarily recommending you do this, just want to get your thoughts.

kjgalvan:
Does this change fall in line with what you were thinking? I know it looks a bit silly to have the send events look so similar for the first handful, but it helps us work separately having a single module for sending the standard events from :)

PR #26, opened 2020-06-17
Add number to shopping cart

kjgalvan:
Adds number of items in shopping cart to icon

cart

### DIFF
--- a/demo/src/App.js
+++ b/demo/src/App.js
@@ -1,15 +1,36 @@
-import React from 'react';
-import {Navbar, Nav, Image} from 'react-bootstrap';
+import React, {useMemo} from 'react';
+import {connect} from 'react-redux';
+import PropTypes from 'prop-types';
+import {Navbar, Nav, Image, Badge} from 'react-bootstrap';
 import {Link} from 'react-router-dom';
 import {Router} from './Router.js';
 import Logo from './logo.svg';
 import './App.css';
 
 /**
- * @return {!JSX} Wrapper component that wraps the entire site with
- *     common components such as a header, navigation, and shopping cart.
+ * Computes the total number of items in the cart.
+ * @param {ItemStore} items A {@link ItemStore} listing of items
+ * to purchase and their desired quantity.
+ * @return {number} The total quantity.
  */
-export function App() {
+function computeNumberOfItemsInCart(items) {
+  let totalItems = 0;
+  for (const item of Object.values(items)) {
+    totalItems += item.quantity;
+  }
+  return totalItems;
+}
+
+/**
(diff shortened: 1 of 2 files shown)


demo/src/App.js
```diff
@@ -1,15 +1,36 @@
...
 import {Router} from './Router.js';
 import Logo from './logo.svg';
 import './App.css';
 
 /**
- * @return {!JSX} Wrapper component that wraps the entire site with
- *     common components such as a header, navigation, and shopping cart.
+ * Computes the total number of items in the cart.
+ * @param {ItemStore} items A {@link ItemStore} listing of items
+ * to purchase and their desired quantity.
```
calhounsm, line 13: Nit: Indent 4 spaces for continuation lines on @param documentation.
https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments

demo/src/App.js
```diff
@@ -1,15 +1,36 @@
...
+  for (const item of Object.values(items)) {
+    totalItems += item.quantity;
+  }
+  return totalItems;
+}
+
+/**
+ * Creates wrapper component that wraps the entire site with common components
+ * such as a header, navigation, and shopping cart.
+ * @param {!ItemStore} items The items stored in the site state
```
calhounsm, line 27: Nit: Trailing period.

demo/src/App.js
```diff
@@ -1,15 +1,36 @@
...
+ * @return {!JSX}
+ */
+const AppBase = function({items}) {
+  const itemsInCart = useMemo(() => {
+    return computeNumberOfItemsInCart(items);
+  }, [items]);
```
calhounsm, line 33: Just want to confirm due to my lack of familiarity with react.

The useMemo function here is saying that the first time the page is rendered, compute the number of items in the cart.  Then on each subsequent render, only recompute this value if the items attribute of the App has changed, i.e. the contents of the cart is different.

Is that about correct?

demo/src/App.js
```diff
@@ -33,10 +54,27 @@ export function App() {
...
-}
+};
+
+AppBase.propTypes = {
+  items: PropTypes.object,
+};
+
+const mapStateToProps = (state) => ({
+  items: state.items,
+});
```
calhounsm, line 78: Also for my understanding: this + the connect below is saying that rather than reading items from the HTML attribute items, it should come from the app's state.items JS property?

demo/src/App.js
kjgalvan: Yes, that is correct :)

demo/src/App.js
kjgalvan: Yes this is to get the items from the state

PR #27, opened 2020-06-17
Deploy app to the world wide web

Alex7Li:
I set up the demo folder so that we can easily deploy it to google app engine in the future. One noteworthy change for @kjgalvan  @pedrosilva0111 is that react start now runs the build directly, so you have to call yarn build before running yarn start. I also added a command yarn qstart that does the same thing as yarn start used to do. This behavior is documented in the readme. 

I deployed the current version of the site here: http://measurement-library.appspot.com/

Deployment instructions for future reference:
First, download gcloud at https://cloud.google.com/sdk/docs and open the measurement-library project from your corp account (you have to choose out of like 1500). Next, navigate to the demo folder of measurement library and run yarn build, then gcloud app deploy. Then you can go to the site, or run gloud app browse to see it.

EDIT: Looking into editing this commit so as not to break yarn start by request of KJ

### DIFF
--- a/demo/README.md
+++ b/demo/README.md
@@ -4,8 +4,7 @@ The demo website to show features of the measure library.
 This site fires actual gtag and measure library events, and
 displays the code for them.
 
-[To run the site, click here!](https://measurement-library.wl.r.appspot.com/)
-
+[To run the site, click here!](http://measurement-library.appspot.com/)
 
 # Deploying the app locally
 
--- a/demo/README.md
+++ b/demo/README.md
@@ -1,0 +1,31 @@
+# Demo Website
+
+The demo website to show features of the measure library.
+This site fires actual gtag and measure library events, and
+displays the code for them.
+
+[To run the site, click here!](https://measurement-library.wl.r.appspot.com/)
+
+
+# Deploying the app locally
+
+To run the app locally, navigate to this /demo directory,
+then type the following commands in the console.
+
+```shell script
+yarn build
+yarn start
+```
+
+For a faster startup without building any files,
+you can run the command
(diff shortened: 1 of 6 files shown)


demo/README.md
```diff
@@ -0,0 +1,30 @@
+# Demo Website
+
+The demo website to show features of the measure library.
+This site fires actual gtag and measure library events, and
+displays the code for them.
+
+[To run the site, click here!](http://measurement-library.appspot.com/)
```
zikas, line 7: I'm not sure how long we'll maintain this appengine up and running and I'm concerned that this statement might become outdated.. 
Please consider documenting only the instruction about how to run your own instance (local or a command to deploy to the app engine) and let's keep this demo side for ourselves.

demo/server.js
```diff
@@ -0,0 +1,17 @@
+const express = require('express');
+const app = express();
+const path = require('path');
+
+// Serve the static files from the React app
+app.use(express.static(path.join(__dirname, 'build')));
+
+
```
zikas, line 8: nit: extra blank line

Having multiple blank lines is not encouraged
https://google.github.io/styleguide/jsguide.html#formatting-vertical-whitespace

demo/server.js
```diff
@@ -0,0 +1,17 @@
...
+app.get('*', (req, res) => {
+  res.sendFile(path.join(__dirname+'/build/index.html'));
```
zikas, line 10: `__dirname + '/build/index.html'`

https://google.github.io/styleguide/jsguide.html#formatting-horizontal-whitespace

PR #33, opened 2020-06-18
Thanks screen

pedrosilva0111:
Created the thanks screen, fully functional.

### DIFF
--- a/demo/src/screens/ThankScreen/ThankScreen.js
+++ b/demo/src/screens/ThankScreen/ThankScreen.js
@@ -5,7 +5,7 @@ import {connect} from 'react-redux';
 import PropTypes from 'prop-types';
 
 /**
- * @param {{items:Array}} items The items stored in the site state
+ * @param {{items:Array}} items The items stored in the site state.
  * @return {!JSX} Page component where we thank the user
  *     for their purchase(s).
  */
--- a/demo/src/screens/ThankScreen/ThankScreen.js
+++ b/demo/src/screens/ThankScreen/ThankScreen.js
@@ -5,7 +5,7 @@ import {connect} from 'react-redux';
 import PropTypes from 'prop-types';
 
 /**
- * @param {!Object} items The items stored in the site state
+ * @param {{items:Array}} items The items stored in the site state
  * @return {!JSX} Page component where we thank the user
  *     for their purchase(s).
  */
@@ -15,18 +15,16 @@ export const ThankScreenBase = function({items}) {
     const products = [];
     const usedItemSet = new Set();
 
-    let i;
-    for (i = 0; i < ITEMS_TO_DISPLAY; ++i) {
-      let maxPrice = 0;
+    for (let i = 0; i < ITEMS_TO_DISPLAY; ++i) {
       let maxPriceItem;
       let maxPriceID;
 
       // Finds and pushes the most expensive item that has not yet been used.
       for (const [itemID, item] of Object.entries(items)) {
-        if (item.cost > maxPrice && !usedItemSet.has(itemID)) {
+        if (item.cost > (maxPriceItem ? maxPriceItem.cost: 0) &&
+            !usedItemSet.has(itemID)) {
           maxPriceItem = item;
           maxPriceID = itemID;
-          maxPrice = item.cost;
         }
       }
       products.push(
@@ -53,9 +51,9 @@ export const ThankScreenBase = function({items}) {
         <Col xs={10}>
           <Jumbotron>
             <div style = {{paddingBottom: 50, fontSize: 30}}>
-              {'Thank you for your purchase!'}
+              Thank you for your purchase!
             </div>
-            {'Here are some other items you may be interested in:'}
+            Here are some other items you may be interested in:
             <br/>
             <ul className="product-list">
               {productListings}
@@ -81,7 +79,7 @@ const mapStateToProps = (state) => ({
 });
 
 /*
- * Decorate the ProductScreenBase function, implicity passing in
+ * Decorate the ThankScreenBase function, implicity passing in
  * information about the global site state. Also pass mapDispatchToProps
  * function to allow modification of global cart.
  */
--- a/demo/src/screens/ThankScreen/ThankScreen.js
+++ b/demo/src/screens/ThankScreen/ThankScreen.js
@@ -5,20 +5,23 @@ import {connect} from 'react-redux';
 import PropTypes from 'prop-types';
 
 /**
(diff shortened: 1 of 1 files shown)


pedrosilva0111:
Screenshot 2020-06-18 at 2 49 14 PM

demo/src/screens/ThankScreen/ThankScreen.js
```diff
@@ -1,13 +1,88 @@
-import React from 'react';
+import React, {useMemo} from 'react';
+import {Container, Row, Col, Jumbotron} from 'react-bootstrap';
+import {ProductListing} from '../../components/ProductListing/ProductListing.js';
+import {connect} from 'react-redux';
+import PropTypes from 'prop-types';
 
 /**
+ * @param {!Object} items The items stored in the site state
```
zikas, line 8: Should be 
`@param {{items: Array}} items ...`

per https://google.github.io/styleguide/jsguide.html#features-objects-destructuring


Also, it's practice to end descriptions with a dot.

demo/src/screens/ThankScreen/ThankScreen.js
```diff
@@ -1,13 +1,88 @@
...
+      let maxPrice = 0;
+      let maxPriceItem;
+      let maxPriceID;
+
+      // Finds and pushes the most expensive item that has not yet been used.
+      for (const [itemID, item] of Object.entries(items)) {
+        if (item.cost > maxPrice && !usedItemSet.has(itemID)) {
+          maxPriceItem = item;
+          maxPriceID = itemID;
+          maxPrice = item.cost;
```
zikas, line 29: seems redundant, it should be sufficient to just maxPriceItem.cost (similar to how it's done for maxPriceItem.name). If it can be undefined you could always `maxPriceItem.cost || 0`.

demo/src/screens/ThankScreen/ThankScreen.js
```diff
@@ -1,13 +1,88 @@
...
-    <div>
-      <h1>{`I'm the Thanks Page!`}</h1>
-    </div>
+    <Container>
+      <Row>
+        <Col/>
+        <Col xs={10}>
+          <Jumbotron>
+            <div style = {{paddingBottom: 50, fontSize: 30}}>
+              {'Thank you for your purchase!'}
```
zikas, line 56: I'm not sure what is the value in the literals being wrapped with curly braces. 

Examples I see just use plain text 
https://reactjs.org/docs/introducing-jsx.html#specifying-children-with-jsx

demo/src/screens/ThankScreen/ThankScreen.js
```diff
@@ -1,13 +1,88 @@
...
+
+ThankScreenBase.propTypes = {
+  items: PropTypes.object,
+};
+
+/**
+  * @param {{items:Itemstore}} state
+  * @return {{items:Itemstore}} The items stored in the site state.
+  */
+const mapStateToProps = (state) => ({
```
zikas, line 79: I'm confused about what this method does? How is it different from just returning the state as-is or `Object.assign({}, state);` if the intention is to make a copy.

demo/src/screens/ThankScreen/ThankScreen.js
```diff
@@ -1,13 +1,88 @@
...
  *     for their purchase(s).
  */
-export function ThankScreen() {
+export const ThankScreenBase = function({items}) {
+  const /** Array<!JSX> */ productListings = useMemo(() => {
+    const ITEMS_TO_DISPLAY = 6;
+    const products = [];
+    const usedItemSet = new Set();
+
+    let i;
```
zikas, line 18: it seems like it should be a part of `for` loop definition
```
for (let i = 0; i < ITEMS_TO_DISPLAY; ++i) {...}
```

demo/src/screens/ThankScreen/ThankScreen.js
```diff
@@ -1,13 +1,88 @@
...
+/**
+  * @param {{items:Itemstore}} state
+  * @return {{items:Itemstore}} The items stored in the site state.
+  */
+const mapStateToProps = (state) => ({
+  items: state.items,
+});
+
+/*
+ * Decorate the ProductScreenBase function, implicity passing in
```
zikas, line 84: ThankScreenBase

demo/src/screens/ThankScreen/ThankScreen.js
pedrosilva0111: I am also still a bit confused by this, but The react-redux connect function takes a mapStateToProps function so that it can merge the selected parts of the redux state with the component's props

zikas: Won't suggested `Object.assign({}, state)` solve this problem? (the advantage is we reduce the maintenance burden of cherry-picking properties)

demo/src/screens/ThankScreen/ThankScreen.js
```diff
@@ -15,18 +15,16 @@ export const ThankScreenBase = function({items}) {
...
-    for (i = 0; i < ITEMS_TO_DISPLAY; ++i) {
-      let maxPrice = 0;
+    for (let i = 0; i < ITEMS_TO_DISPLAY; ++i) {
       let maxPriceItem;
       let maxPriceID;
 
       // Finds and pushes the most expensive item that has not yet been used.
       for (const [itemID, item] of Object.entries(items)) {
-        if (item.cost > maxPrice && !usedItemSet.has(itemID)) {
+        if (item.cost > (maxPriceItem ? maxPriceItem.cost: 0) &&
```
zikas, line 24: nit: missing space `maxPriceItem.cost : 0`

demo/src/screens/ThankScreen/ThankScreen.js
pedrosilva0111: Ok so I talked to KJ since she has worked with react quite a bit. We went over the react redux documentation and came up with a conclusion for why mapStateToProps is being used this way. We aren't actually making a copy, we are cherry picking properties of state because mapStateToProps passes the property as a React prop to the ThanksScreenBase component. If we pass in the entire state, the component will re-render for every state change instead of only re-rendering when the items object is changed. Since this component is only using a a specific part of the state, it will only re-render when that part changes. Here is the documentation we used https://react-redux.js.org/using-react-redux/connect-mapstate#return !

zikas: Ah, this makes sense now.
Thank you for looking into this!

PR #34, opened 2020-06-18
Install karma and jasmine testing framework

Alex7Li:
This commit sets up jasmine testing with karma for the main repository (not the demo app!). This will be very useful for when we start writing measurement library code later. Added instructions for installation in the README.

Screenshot 2020-06-18 at 5 32 33 PM


spec/example.spec.js
```diff
@@ -0,0 +1,15 @@
+describe('Test that the tests are running', function(){
```
durbanie, line 1: What is the plan for this? Will this file live on forever, or is it only an example? I think an example is a fine idea (it will give me something to build against for Travis), but add a TODO to remove it once the real code/tests are added.

durbanie, line 1: If we still want to keep this file, I'd do two things:

1. Put it in a "test" folder, rather than "spec". 

2. I'd like to stick with the Google convention, which is to name test files the same as their source, with a "_test" subscript (e.g. my-source-file.js -> my-source-file_test.js).

Alex7Li: I was planning to delete it later, I'll add a TODO

Alex7Li: Ok, I changed the name to fit the convention

PR #37, opened 2020-06-22
Clear cart after checkout

kjgalvan:
Removes all items from cart when checkout is complete.

### DIFF
--- a/demo/src/screens/CheckoutScreen/CheckoutScreen.js
+++ b/demo/src/screens/CheckoutScreen/CheckoutScreen.js
@@ -28,7 +28,7 @@ const BILLING_FORM_ID = 'billing-info-form';
  * @param {function()} clearCart A function to clear all items from cart.
  * @return {!JSX}
  */
-const CheckoutScreenBase = function({clearCart}) {
+const CheckoutScreenBase = ({clearCart}) => {
   const [shippingDone, setShippingDone] = useState(false);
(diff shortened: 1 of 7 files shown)


demo/src/screens/CheckoutScreen/CheckoutScreen.js
```diff
@@ -22,9 +25,10 @@ const BILLING_FORM_ID = 'billing-info-form';
 /**
  * Page component for a user to enter in personal billing information
  * and confirm the items in their cart.
+ * @param {function()} clearCart A function to clear all items from cart.
  * @return {!JSX}
  */
-export function CheckoutScreen() {
+const CheckoutScreenBase = function({clearCart}) {
```
durbanie, line 31: Nit: Can you make this an arrow function? This is usually preferred over the function keyword in the style guide:
https://google.github.io/styleguide/jsguide.html#features-functions-arrow-functions

kjgalvan: Done! I also changed this pattern elsewhere and did some manual testing to ensure it didn't break anything. Thanks for the suggestion.

durbanie: Nice, thanks!

kjgalvan:
@durbanie It's ready to merge now. Thanks!

PR #29, opened 2020-06-18
Add gtag to carousel

kjgalvan:
Sends promotion data to Google Analytics for the home carousel.

PR #39, opened 2020-06-22
Add code modals to carousel

kjgalvan:
NOTE: This is built on top of PR #29 and the diff will look incorrect until it is merged. View commit [757e4cc](https://github.com/googleinterns/measurement-library/commit/757e4cca45e7a4861f000283b8894acb92ddfbc4) to see correct code changes.

Adds code modals to the Home Carousel.

image

### DIFF
--- a/demo/src/screens/HomeScreen/HomeCarousel/HomeCarousel.js
+++ b/demo/src/screens/HomeScreen/HomeCarousel/HomeCarousel.js
@@ -30,7 +30,6 @@ export function HomeCarousel() {
         promotion_name: 'Home Carousel',
       },
     ],
-    promotion_name: 'Home Carousel',
   };
 
   // Send view_promotion event when the carousel loads for the first time.
--- a/demo/src/screens/HomeScreen/HomeCarousel/HomeCarousel.js
+++ b/demo/src/screens/HomeScreen/HomeCarousel/HomeCarousel.js
@@ -111,7 +111,7 @@ export function HomeCarousel() {
         <Carousel.Caption>
           <div className="header-with-modal carousel-header">
             <h3>
-              Stunning High Quality Prints Available
+              {`Stunning High Quality Prints Available`}
               {carouselCodeModal}
             </h3>
           </div>
@@ -123,7 +123,7 @@ export function HomeCarousel() {
                 }}
                 to="/product/7ba94"
               >
-              latest piece
+                {`latest piece`}
               </Link>
             </p>
             {getSelectPromotionCodeModal('7ba94')}
@@ -139,7 +139,7 @@ export function HomeCarousel() {
         <Carousel.Caption>
           <div className="header-with-modal carousel-header">
             <h3>
-              SALE!
+              {`SALE!`}
             </h3>
             {carouselCodeModal}
           </div>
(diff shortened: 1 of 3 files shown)


demo/src/screens/HomeScreen/HomeCarousel/HomeCarousel.js
```diff
@@ -1,16 +1,106 @@
...
+   *     an items attibute mapped to an array of {@link ItemParameters}
+   */
+  function getSelectPromotionParameters(itemId, discount = 0) {
+    return {
+      items: [
+        {
+          ...getItemParameters(itemId),
+          promotion_name: 'Home Carousel',
+          // if discount is positive, include it in the item parameter
+          ...(discount > 0 && {discount}),
```
zikas, line 66: Although it seems like a convenient shortcut, it requires an assumption that ...(false) behaves correctly on all browsers. 
So just to be extra cautious and don't require from a reader to go over edge-cases in the specification, I'd recommend using `discount > 0 ? {discount} : {}`

demo/src/screens/HomeScreen/HomeCarousel/HomeCarousel.js
```diff
@@ -46,10 +166,26 @@ export function HomeCarousel() {
...
-          </p>
+          <div className="header-with-modal carousel-header">
+            <h3 className="font-italic">
+              Toe Beans
+            </h3>
+            {carouselCodeModal}
+          </div>
+          <div className="header-with-modal carousel-text">
+            <p>
+              {`Seen like never before. `}
```
zikas, line 177: Why "Toe Beans" is used without {} but this line requires wrapping?

I'd recommend being consistent with across the codebase and probably establish general rules where one or the other style should be used and agree with all team members to follow it.

kjgalvan: I usually use template strings in JSX to preserve whitespace as a lot of code editors will auto trim these on save. This one has a trailing space that is necessary whereas Toe Beans does not. I am fine using them all over to keep things consistent and future-proof though so I will make the appropriate edits.

demo/src/screens/HomeScreen/HomeCarousel/HomeCarousel.js
kjgalvan: This is something I hadn't considered before - thanks for the suggestion!

demo/src/screens/HomeScreen/HomeCarousel/HomeCarousel.js
```diff
@@ -1,16 +1,106 @@
...
+        ...getItemParameters('hjdf7'),
+        promotion_name: 'Home Carousel',
+        discount: 25,
+      },
+      {
+        ...getItemParameters('3h488'),
+        promotion_name: 'Home Carousel',
+      },
+    ],
+    promotion_name: 'Home Carousel',
```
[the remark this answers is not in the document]
kjgalvan: It may be outdated / a case of incorrect reading of the docs. I used the event reference to select parameters and promotion_name is both at the top level and item level:

https://developers.google.com/gtagjs/reference/aw-events#view_promotion

PR #41, opened 2020-07-01
Set up Travis CI to run tests.

durbanie:
This is a Travis "MVP". By adding .travis.yml (along with the existing yarn.lock), Travis will run "yarn install" followed by "yarn test".

This PR also adds the "test" invocation in package.json. For now, it runs just the example tests that Alex set up in headless chrome.

The results of my latest run are here: https://travis-ci.com/github/durbanie/measurement-library

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,16 @@ yarn install
 npm install
 ```
 
-Next, add karma to your path if it isn't already.
+Next, run the tests using yarn or npm:
+
+```shell script
+yarn test
+# or
+npm test
+```
+
+## Running tests interactively
+You can also run the tests "interactively" via Karma directly. Add Karma to your path if it isn't already:
 
 ```shell script
 yarn global add karma-cli
@@ -22,7 +31,7 @@ npm install -g karma-cli
 ```
 
 Next, run the tests with karma
+
 ```shell script
 karma start
 ```
-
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -13,7 +13,16 @@ yarn install
...
+Next, run the tests using yarn or npm:
+
+```shell script
+yarn test
+# or
+npm test
+```
+
+## Running tests interactively
+You can also run the tests "interactively" via Karma directly. Add Karma to your path if it isn't already:
```
Alex7Li, line 25: When you call yarn test, it calls karma. Does that step require karma to be globally installed? If so, it might be better to move the karma path line above

durbanie: Nope, karma-cli is not required to be globally installed when run via yarn or npm. I have tested this on a machine where I don't have it installed, and both "yarn test" and "npm test" work as expected.

Thanks for bringing this up though!

package.json
```diff
@@ -9,5 +9,8 @@
     "karma-jasmine": "^3.3.1",
     "karma-jasmine-html-reporter": "^1.5.4",
     "karma-spec-reporter": "^0.0.32"
+  },
+  "scripts": {
+    "test": "karma start --single-run --browsers ChromeHeadless"
```
calhounsm, line 14: It'd probably be nice to add a couple additional browsers here, but starting with just Chrome SGTM.

PR #42, opened 2020-07-09
Set up travis to build and run tests

Alex7Li:
This commit more or less copies the build system from the https://github.com/google/data-layer-helper repository to here. The goal is to set up continuous integration for future commits to the repository, ensuring that we don't commit code that doesn't compile or fails tests. This commit uses grunt to do all the work, though since commands are called indirectly through yarn we don't require any global installs for either grunt or karma.


Gruntfile.js
```diff
@@ -0,0 +1,93 @@
...
+        files: {
+          'dist/measure.js': 'src/**/*.js',
+        },
+        options: {
+          js: [
+            'node_modules/google-closure-library/closure/goog/base.js'
+          ],
+          hide_warnings_for: 'google-closure-library',
+          compilation_level: 'ADVANCED_OPTIMIZATIONS',
+          language_in: 'ECMASCRIPT6',
```
kjgalvan, line 35: Could you add an explicit language_out flag?

package.json
```diff
@@ -1,16 +1,29 @@
...
-    "karma-chrome-launcher": "^3.1.0",
-    "karma-jasmine": "^3.3.1",
-    "karma-jasmine-html-reporter": "^1.5.4",
-    "karma-spec-reporter": "^0.0.32"
-  },
-  "scripts": {
-    "test": "karma start --single-run --browsers ChromeHeadless"
+    "grunt": "~1.2.1",
+    "grunt-karma": "~4.0.0",
+
```
kjgalvan, line 13: I don't think spacing out the dependencies is standard / recommended

test/integration/integration_test.js
```diff
@@ -0,0 +1,8 @@
+/*
+ * TODO: Replace this test file with real tests
+ */
+describe('Test that the integration tests are running', function(){
+  it('should pass this test', function(){
```
kjgalvan, line 5: Could you change these to arrow functions?

test/integration/karma.conf.js
```diff
@@ -0,0 +1,11 @@
...
+
+module.exports = function(config) {
+  baseConfig(config);
+  config.set({
+    files: [
+      './**_test.js',
+      '../../dist/measure.js',
+    ],
+  });
+};
```
kjgalvan, line 11: I believe according to the style guide a single blank line should follow the file's implementation: https://google.github.io/styleguide/jsguide.html#source-file-structure

Alex7Li:
For some reason the build was marked as pending for several hours for me despite the actual travis build being complete. Maybe there was an issue with sending this to github; I'm not sure exactly why that happened but it doesn't seem to have anything to do with my code, I went to the travis CI and restarted the build, it worked the second time. Worth mentioning since this is supposed to be the travis PR

.travis.yml
```diff
@@ -1,3 +1,13 @@
 language: node_js
 node_js:
   - "14"
+addons:
+  firefox: latest
```
zikas, line 5: out of curiosity, why ff is latest but chrome is stable?

Gruntfile.js
```diff
@@ -0,0 +1,95 @@
...
+ * limitations under the License.
+ */
+
+module.exports = function(grunt) {
+  const closurePackage = require('google-closure-compiler');
+  closurePackage.grunt(grunt);
+
+  grunt.initConfig({
+    'pkg': grunt.file.readJSON('package.json'),
+
```
zikas, line 23: nit: please remove extra blank line

.travis.yml
Alex7Li: The docs for travis don't have any version of firefox named 'stable' https://docs.travis-ci.com/user/firefox or any version of chrome named 'latest' https://docs.travis-ci.com/user/chrome#selecting-a-chrome-version. I think it just mimics what chrome/firefox name files on their site, rather than having a consistent API.

PR #43, opened 2020-07-22
Minimal setup neccessary for a passing unit test

Alex7Li:
This commit bootstraps the measurement library, getting the compiler, linter, and unit tests to run correctly on the core source code. Really the linter should have been done in a previous commit but I forgot about it. We need the data-layer-helper repo as a primary dependency. It still will take some work before that gets to version 1.0.0 so until then, it is a git submodule.

The measurement library is setup to work according to https://docs.google.com/document/d/18GIuwxgZjU0rOvUDBjiEUpJje_BPFgydNl7sIWnXjZ4/edit#


src/measure.js
```diff
@@ -0,0 +1,26 @@
...
+   * @const {!DataLayerHelper} The DataLayerHelper to use with this application.
+   */
+  const helper = new DataLayerHelper(window.dataLayer);
+
+  // TODO
+}
+
+setup();
+
+exports = {setup};
```
pedrosilva0111, line 26: I think it would be nice to have a new line at the end here

test/config_test.js
```diff
@@ -0,0 +1,72 @@
...
+
+const {setup} = goog.require('measurementlibrary.measure');
+
+let storageInterface;
+let eventProcessor;
+let measure;
+
+/**
+ * Run a test, once assuming that the measure snippet fired before the call
+ * to setup, then again assuming that the gtag snippet fired after the call
```
pedrosilva0111, line 12: Should this be the measure snippet?

test/config_test.js
```diff
@@ -0,0 +1,72 @@
...
+ * to setup.
+ * @param {function} config The configuration done in the gtag code snippet
```
pedrosilva0111, line 14: I may be missing something, but I think this might also say gtag on accident :)

test/config_test.js
Alex7Li: Yep

src/measure.js
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Listen to events passed to the current window.dataLayer (or a new one
+ * if none exist).
+ */
+function setup() {
+  /**
+   * @global {!Array} The data layer of the application, storing a record of
+   * events that have changed over time.
+   */
+  window.dataLayer = window.dataLayer || [];
```
durbanie, line 14: Does the dataLayer always need to be named "dataLayer"? If not, maybe add it as a parameter to "setup" and pass it in?

karma.conf.js
```diff
@@ -8,10 +8,12 @@ module.exports = function(config) {
     // Automatically run tests for files matching these regex.
     files: [
       // ----------------- Third Party Dependencies ----------------------------
-      {pattern: 'node_modules/google-closure-library/closure/goog/base.js'},
-      // ------------------------ Source Files ---------------------------------
       // Dependencies must be listed before the file they are used in for
       // the googmodule preprocessor to function properly.
+      {pattern: 'node_modules/google-closure-library/closure/goog/base.js'},
+      {pattern: 'data-layer-helper/src/plain/is_plain_object.js'},
```
durbanie, line 14: Why not use wildcards here?

test/config_test.js
```diff
@@ -0,0 +1,73 @@
+goog.module('measurementlibrary.measure.testing.setup');
```
durbanie, line 1: I'm not sure I like 'config' for the name of the file. Why not 'setup'?

karma.conf.js
Alex7Li: Otherwise, the loading order is wrong and the googmodule preprocessor doesn't work

test/config_test.js
Alex7Li: Fixed, I had renamed the function before

src/measure.js
Alex7Li: I think it will almost always be named dataLayer, but maybe there is another use case I haven't thought of so I will add the default parameter

test/utils.js
```diff
@@ -0,0 +1,41 @@
...
+ * Run a test, once assuming that the measure snippet fired before the call
+ * to setup, then again assuming that the measure snippet fired after the call
+ * to setup.
+ * @param {function(!Array<*>):undefined} config The configuration
+ *     to be done in the measure code snippet.
+ * @param {function(!Array<*>):undefined} test The tests to run
+ *     after both the code snippet and setupMeasure function have fired.
+ * @param {function():undefined} reset A function to reset state between tests.
+ */
+function runInBothOrders(config, test, reset = noop) {
```
durbanie, line 18: I'm not sure about this pattern. Considering race conditions is commendable, but there are a number of potential issues here:

1. There's a lot of indirection, which can obscure what the test is actually doing. At a minimum, new developers/users of the code will need to switch back and forth between this helper file and the actual test code to understand what's going on. If this will be only used for setup unit tests, I'd move the helper function to that file. I'd strongly recommend against using this plan for all tests.

2. Jasmine has well-accepted ways of resetting state (beforeAll/beforeEach), and this jumps out of that usual flow. 

3. I'm not sure it will be clear to the user exactly where the test fails. Wrapping each separate case in a "describe" or "it" statement could help (I think Jasmine supports multiple nested describes, but not sure about its).

4. IF we decide to keep this method of testing (which I'd recommend against), I think we need a more descriptive name of the helper function.

5. On the topic of testing for race conditions in general, I'd consider whether or not every case merits a unit test: 
- Are there particular features potentially sensitive to race conditions? Those might be worth testing in a unit test.
- Is there something about the library as a whole that is sensitive to race conditions? In that case, it should probably be addressed in integration tests. Yes, the snippet of code which is responsible for ensuring correct behavior when setup/snippet order changes should be covered in a handful of unit tests, but we definitely don't need to do that in every test.



In general, I'd prefer to think of tests as documentation for how the code should work. If a new user wants to understand how the library works, how difficult will it be for them to learn about it from the tests alone? If tests are documentation, then generally they should be more descriptive. This is a somewhat controversial topic in the developer world at large, but Google tends to prefer it this way: https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html

Alex7Li: Discussed in person
1, 5) Moved the function to just the setup unit tests.
2, 3) Apparently you can nest if blocks inside of functions nested inside of describe blocks!

src/setup.js
```diff
@@ -0,0 +1,22 @@
+goog.module('measurementLibrary.setup');
```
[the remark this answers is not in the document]
Alex7Li: Done

test/setupMeasure_test.js
```diff
@@ -0,0 +1,111 @@
...
+ *     to be done in the measure code snippet.
+ * @param {function(!Array<*>):undefined} test The tests to run
+ *     after both the code snippet and setupMeasure function have fired.
+ */
+function executeSnippetBeforeAndAfterSetup(config, test) {
+  let snippet;
+  let dataLayer;
+  beforeEach(() => {
+    dataLayer = [];
+    // The code snippet that is run asynchronously.
```
durbanie, line 23: Can you add some additional context as to what this is doing? It looks to me like this is just setting up a fake version of "measure". Is this temporary? If so, can you add a "TODO"?

Alex7Li: This is not temporary, the html file that calls measurement library has two components - the script tag that calls the library and the script tag that defines user settings. The code in this before each is the latter script tag.

test/setupMeasure_test.js
```diff
@@ -0,0 +1,111 @@
...
+    snippet = (dataLayer) => {
+      const measure = function() {
+        dataLayer.push(arguments);
+      };
+      config(measure);
+    };
+  });
+  it('runs snippet first', () => {
```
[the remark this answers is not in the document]
Alex7Li: Done, though now the sentence doesn't read like it should start with 'describe' or 'it'. I think it's not very feasible with the way I'm using jasmine in this file

test/setupMeasure_test.js
durbanie: Sounds good.

PR #44, opened 2020-07-23
Create interfaces for eventProcessor and Storage

Alex7Li:
This commit creates interfaces for the eventProcessor and Storage. This is important since this library will potentially support several different kinds of eventProcessor/Storage classes and this provides a foundation that can be built off of and referenced in the future. Also it helps with jsdoc.

One change I wasn't sure about was documenting the processEvent function of EventProcessor. It uses rest parameters for more clarity. I think that no polyfill will happen since this is just an interface, but it may be worth checking out the dist size once we have an implementation to make sure that this is true.

### DIFF
--- a/src/EventProcessor/EventProcessorInterface.js
+++ b/src/EventProcessor/EventProcessorInterface.js
@@ -20,7 +20,7 @@ class EventProcessor {
    * @param {{get:function(string):*, set:function(string, *)}} modelInterface
    *    An interface to load or save short term page data from the data layer.
    * @param {string} eventName The name of the event passed to the data layer.
-   * @param {...*} eventArgs The events passed to the data layer
+   * @param {Object<string, *>} eventArgs The events passed to the data layer
    */
   processEvent(storageInterface, modelInterface, eventName, ...eventArgs) {};
 
--- a/src/EventProcessor/EventProcessorInterface.js
+++ b/src/EventProcessor/EventProcessorInterface.js
@@ -13,7 +13,7 @@ goog.module('measurementlibrary.EventProcessor.EventProcessorInterface');
  */
 class EventProcessor {
   /**
-   * Process an event passed into the data layer.
+   * Processes an event passed into the data layer.
    *
    * @param {!StorageInterface} storageInterface An interface to an object to
    *    load or save persistent data with.
@@ -25,7 +25,7 @@ class EventProcessor {
   processEvent(storageInterface, modelInterface, eventName, ...eventArgs) {};
 
   /**
-   * Decide if a given key should be persisted to long term storage, or just
+   * Decides if a given key should be persisted to long term storage, or just
    * saved locally to the data layer.
    *
    * @param {string} key The location at which to store data.
--- a/src/EventProcessor/EventProcessorInterface.js
+++ b/src/EventProcessor/EventProcessorInterface.js
@@ -22,7 +22,7 @@ class EventProcessor {
    * @param {string} eventName The name of the event passed to the data layer.
    * @param {Object<string, *>} eventArgs The events passed to the data layer
    */
-  processEvent(storageInterface, modelInterface, eventName, ...eventArgs) {};
+  processEvent(storageInterface, modelInterface, eventName, eventArgs) {};
 
   /**
    * Decides how long a given key should be persisted to long term storage for.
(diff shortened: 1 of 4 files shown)


src/EventProcessor/EventProcessorInterface.js
```diff
@@ -0,0 +1,39 @@
...
+ * Event processors may perform actions such as sending a network request
+ * based on the event data, interacting with the storage interface to commit
+ * and retrieve values from long-term storage, or reading values from
+ * the abstract model.
+ *
+ * @interface
+ */
+class EventProcessor {
+  /**
+   * Process an event passed into the data layer.
```
kjgalvan, line 16: Nit: This is the wrong tense for method description. Applies to rest of PR as well.

Alex7Li: Good catch, fixed

src/EventProcessor/EventProcessorInterface.js
```diff
@@ -0,0 +1,45 @@
...
+class EventProcessor {
+  /**
+   * Processes an event passed into the data layer.
+   *
+   * @param {!StorageInterface} storageInterface An interface to an object to
+   *    load or save persistent data with.
+   * @param {{get:function(string):*, set:function(string, *)}} modelInterface
+   *    An interface to load or save short term page data from the data layer.
+   * @param {string} eventName The name of the event passed to the data layer.
+   * @param {...*} eventArgs The events passed to the data layer
```
calhounsm, line 23: What would an example invocation with event args look like?  I was under the impression that the eventArgs would be an optional object, rather than a varargs.

src/Storage/StorageInterface.js
```diff
@@ -0,0 +1,37 @@
...
+   * @param {*} value The data to store.
+   * @param {number=} secondsToLive The maximum number of seconds that the key
+   *    should be saved for.
+   */
+  save(key, value, secondsToLive) {};
+
+  /**
+   * Loads the value associated with a given key in the storageInterface.
+   *
+   * @param {string} key The location at which to load data.
```
calhounsm, line 25: Looks like key is listed twice in the JSdoc.

src/Storage/StorageInterface.js
```diff
@@ -0,0 +1,37 @@
...
+   * @param {string} key The location at which to store data in the model.
+   *    Dot notation is used to access a nested value (i.e. 'employees.jim'
+   *    is the key 'jim' in the nested 'employees' object).
+   * @param {*=} defaultValue The value to load if no key is stored at
```
calhounsm, line 29: Nit: Value to return, rather than load.

src/EventProcessor/EventProcessorInterface.js
Alex7Li: I have it has a varargs, so implementations of the event interface could use an optional object if they want. But in theory you could invoke 
`measure('event', 'pageview', '6:00pm', 'john')`
and it would call
`processEvent(myStorage, myModel, 'pageview', '6:00pm', 'john')`

src/Storage/StorageInterface.js
Alex7Li: Fixed

src/Storage/StorageInterface.js
Alex7Li: Fixed

src/EventProcessor/EventProcessorInterface.js
Alex7Li: Hmm I just read the project plan and it specified that it should be an options object explicitly. So nevermind my last comment, I edited the jsdoc. Now a call would be measure('event', 'pageview', {time:'6:00pm', name:'john'}) and then processEvent(myStorage, myModel, 'pageview', {time:'6:00pm', name:'john'}) .

src/EventProcessor/EventProcessorInterface.js
```diff
@@ -0,0 +1,45 @@
...
+class EventProcessor {
+  /**
+   * Processes an event passed into the data layer.
+   *
+   * @param {!StorageInterface} storageInterface An interface to an object to
+   *    load or save persistent data with.
+   * @param {{get:function(string):*, set:function(string, *)}} modelInterface
+   *    An interface to load or save short term page data from the data layer.
+   * @param {string} eventName The name of the event passed to the data layer.
+   * @param {Object<string, *>} eventArgs The events passed to the data layer.
```
calhounsm, line 23: Should be `!...=`, right?  That is, a non-null optional object.

PR #46, opened 2020-07-28
Added logging configuration

Alex7Li:
Add the ability to log potential misuses of the library to the console

### DIFF
--- a/src/EventProcessor/EventProcessorInterface.js
+++ b/src/EventProcessor/EventProcessorInterface.js
@@ -1,5 +1,5 @@
-goog.require('measurementlibrary.Storage.StorageInterface');
-goog.module('measurementlibrary.EventProcessor.EventProcessorInterface');
+goog.require('measurement_library.Storage.StorageInterface');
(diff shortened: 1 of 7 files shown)


src/EventProcessor/EventProcessorInterface.js
```diff
@@ -1,5 +1,5 @@
-goog.require('measurementlibrary.Storage.StorageInterface');
-goog.module('measurementlibrary.EventProcessor.EventProcessorInterface');
+goog.require('measurement_library.Storage.StorageInterface');
```
durbanie, line 1: Nit: Conventionally, we use camelCase for module names. I don't think it's specified anywhere in the style guide, but here it is for your reference: https://google.github.io/styleguide/jsguide.html#file-goog-module

Alex7Li: I might be misunderstanding something, but this seems to conflict with what you said earlier
https://github.com/google/data-layer-helper/pull/53#discussion_r460156625

PR #47, opened 2020-07-28
Add externs file for the compiler

Alex7Li:
Ensure that closure compiler does not do away with names that are needed for users using the EventProcessor/Storage interfaces by using an externs file.


Alex7Li:
The @export tag is used to export a function defined inside of the library for use outside of a library. The @externs tag is for calling functions defined outside of the library for use inside of the library. Any @extern function can be an @export function, but 
@export compresses the name of a function to a single letter until it exports it, saving more space. On the other hand, @externs can't ever compress the name of a function since it is defined elsewhere, and so it takes up more space in the library.

In this case, someone could write their own custom event processor, so we have to use the @externs tag to integrate their code with the library.

https://developers.google.com/closure/compiler/docs/api-tutorial3

src/externs.js
```diff
@@ -0,0 +1,12 @@
...
+ * third party code. Closure compiler advanced optimizations needs this file
+ * to know not to minimize these function names.
+ * @externs
+ */
+
+function shouldPersist(storageInterface, key, value) {}
+function processEvent(storageInterface, helperInterface, name, args) {}
+
+function set(key, value) {}
+function get(key, value) {}
```
calhounsm, line 12: I'd prefer defining the type and it's methods as externs, rather than a global extern.  This should make the difference between Closure preserving all symbols named processEvent, and just those that are methods of an EventProcessorInterface.

For example, the way Closure defines externs for the Arguments type:
https://github.com/google/closure-compiler/blob/master/externs/es3.js#L230

src/externs.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * @fileoverview This file declares functions that can be declared by external
+ * third party code. Closure compiler advanced optimizations needs this file
+ * to know not to minimize these function names.
+ * @externs
+ */
+
+function shouldPersist(storageInterface, key, value) {}
```
calhounsm, line 8: Am I forgetting, or was storageInterface was removed from this method?  Practically, I don't think it matters in an externs definition, but ideally it would match the real interface.

src/externs.js
```diff
@@ -0,0 +1,12 @@
...
+function processEvent(storageInterface, helperInterface, name, args) {}
+
+function set(key, value) {}
```
calhounsm, line 11: Likewise, storage.set should have the optional TTL?

src/externs.js
Alex7Li: You aren't forgetting, this file was out of date. Fixed.

src/externs.js
Alex7Li: I copied the idea of the code in that file [in a local branch](https://github.com/Alex7Li/measurement-library/blob/externstest/src/externs.js), but it seems that closure still preserves all symbols rather than just the methods. [This file is compiled wrong](https://github.com/Alex7Li/measurement-library/blob/externstest/src/EventProcessor/EventSubclass.js). I'm not sure if I'm misunderstanding your recommendation or if open source closure works differently than you are expecting.

Alex7Li: This solution will work well enough per offline discussion. Refactored so that we don't declare functions twice.

src/eventProcessor/EventProcessorInterface.js
```diff
@@ -39,7 +41,4 @@ class EventProcessor {
    *     If Number.POSITIVE_INFINITY, data will be stored forever.
    */
   persistTime(key, value) {};
-
 }
-
-exports = EventProcessor;
```
calhounsm, line 45: I think not exporting the types might cause a problem, since IIRC the concrete implementations will need to require them (or at least goog.requireType them) for @interface JSDoc annotations.

That said, I think the API looks good so I'm going to merge this.  Feel free to look into this and address it in a followup PR if necessary.

PR #49, opened 2020-07-31
Add eslint

kjgalvan:
Add eslint using same config from Data Layer Helper


Alex7Li:
I actually went out of scope and added this is a pr that hasn't been merged yet #43 but it's better to have it separate so I'll just get rid of it in that one

PR #52, opened 2020-08-05
Config command base

Alex7Li:
This PR creates handles arguments passed to the base configProcessor command. The arguments are converted into an instance of an event handler or storage interface. They will be used for the event and set commands of future CLs.


src/setup.js
```diff
@@ -27,9 +50,62 @@ function setupMeasure(dataLayer) {
...
+    if (typeof eventProcessor === 'string') {
+      if (!(eventProcessor in DEFAULT_EVENT_PROCESSORS)) {
+        log(`No event processor with name ${eventProcessor} found.`,
+            LogLevel.ERROR);
+        return;
+      }
+      processor =
+          new DEFAULT_EVENT_PROCESSORS[eventProcessor](eventOptions);
+    } else {
+      processor = new eventProcessor(eventOptions);
```
durbanie, line 91: I would instead define a variable "EventProcessorCtor" (or similar) and use this if/else to set it appropriately. Then afterwards just do "processor = EventProcessorCtor(eventOptions)"

src/setup.js
```diff
@@ -27,9 +50,62 @@ function setupMeasure(dataLayer) {
...
+   * @param {!Object<string, *>} eventOptions Options objects to pass to the
+   *    event processor.
+   * @param {!StorageConstructor|string} storageInterface An
+   *     interface to a class that will handle setting
+   *     and loading long term data, or a string representing
+   *     an object of this type.
+   * @param {!Object<string, *>} storageOptions Options objects to pass to the
+   *    storage interface.
+   */
+  function configProcessors(eventProcessor, eventOptions,
```
durbanie, line 76: Prefer arrow functions for nested functions: https://google.github.io/styleguide/jsguide.html#features-functions-arrow-functions

However, I'd lean towards defining configProcessors in a separate module and just using it here. I don't know what the tagging team wants to do, but there may be a desire for some pubs to compile this library directly into their code. Having a separate module for each of these may make adapting this library easier for that use case.

src/setup.js
```diff
@@ -27,9 +50,62 @@ function setupMeasure(dataLayer) {
...
+        return;
+      }
+      storage =
+          new DEFAULT_STORAGE_INTERFACES[storageInterface](storageOptions);
+    } else {
+      storage = new storageInterface(storageOptions);
+    }
 
-   // TODO: Use the variables declared above
+    /* TODO add event/set processing
```
durbanie, line 105: I'm not sure about adding all of this code that essentially no-ops. I think this is a good start to this PR and I'm glad we're starting this review, but I think we might need to make sure the additional functionality is in place before merging this.

test/setupMeasure_test.js
```diff
@@ -109,5 +111,16 @@ describe('After calling the setupMeasure function of setup', () => {
...
+    describe('Does not crash when unsupported strings are passed in', () => {
+      executeSnippetBeforeAndAfterSetup(
+          /* config= */
+          (measure) =>
+              measure('config', 'unusedProcessorName', {}, 'unusedStorageName',
+                  {}),
+          /* test=  */ (dataLayer) => {
+            expect(dataLayer.length).toBe(1);
+          });
+    });
```
durbanie, line 123: Anyway we can test what happens if someone passes in something that is not a constructor of one of the interfaces?

src/setup.js
```diff
@@ -15,6 +15,29 @@
 goog.module('measurementLibrary.setup');
 
 const {DataLayerHelper} = goog.require('helper');
+const {LogLevel, log} = goog.require('measurementLibrary.logging');
+
+/**
+ * @typedef {function(new:StorageInterface, !Object<string, *>)}
+ */
+let StorageConstructor;
```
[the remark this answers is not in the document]
Alex7Li: That's probably true. However, I moved the function to a different file and now it's top-level. The type is also used in the DEFAULT_ variables, which is the reason for using a typedef instead of inlining it

src/setup.js
```diff
@@ -27,9 +50,62 @@ function setupMeasure(dataLayer) {
    * The DataLayerHelper to use with this application.
    * @private @const {!DataLayerHelper}
    */
-  const helper = new DataLayerHelper(dataLayer);
+  const helper = new DataLayerHelper(dataLayer, {'processNow': false});
+
+  helper.registerProcessor('config', configProcessors);
+  helper.process();
+
+  /**
```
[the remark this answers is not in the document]
Alex7Li: Originally I was using the full space, but I kept editing it while trying to figure out the jsdoc and whenever I added a word, I would need to edit the next 3 lines. I think that I made my lines too short here but I feel that stylistically, trying to use 'as much of the 80 spaces as possible' is not as good as using around 65 each time. Idk, I filled it up for this PR

src/setup.js
Alex7Li: That's a great idea, done

src/setup.js
Alex7Li: I moved it to a new file

src/setup.js
Alex7Li: Well the additional functionality should be coming in #54 and #55!

test/setupMeasure_test.js
Alex7Li: Added a test for this. I'm not sure if we should let the library crash in these circumstances to fail loudly as soon as possible or just display a error message to the console.

src/setup.js
Alex7Li: I'd prefer not to wait on those two commits before merging, both of them have some template code to mimic what this commit is because they depend on the variables made in this commit. I could have put all 3 in the same CL but it would have gotten quite large

src/config/configProcessors.js
```diff
@@ -0,0 +1,131 @@
...
+ * @param {!StorageConstructor|string} storageInterface An interface to a class
+ *     that will handle setting and loading long term data, or a string
+ *     representing an object of this type.
+ * @param {!Object<string, *>} storageOptions Options objects to pass to the
+ *    storage interface.
+ */
+function configProcessors(helper, eventProcessor, eventOptions,
+                          storageInterface, storageOptions) {
+  // TODO: Read in data from the model as extra options
+  const EventProcessorConstrucutor =
```
durbanie, line 99: Looks like a typo?

src/config/configProcessors.js
```diff
@@ -0,0 +1,131 @@
...
+      getConstructor_(eventProcessor, true);
+  const StorageInterfaceConstructor =
+      getConstructor_(storageInterface, true);
```
durbanie, line 102: Shouldn't this be false?

src/config/configProcessors.js
```diff
@@ -0,0 +1,131 @@
...
+ * DEFAULT_EVENT_PROCESSORS or DEFAULT_STORAGE_INTERFACES, depending on
+ * isEventProcessor. Constructor objects are passed through without
+ * modification.
+ *
+ * @param {!ProcessorConstructor|!StorageConstructor|string} obj
+ * @param {boolean} isEventProcessor True if obj represents an event processor,
+ *     false if obj represents storage
+ * @return {!ProcessorConstructor|!StorageConstructor}
+ */
+function getConstructor_(obj, isEventProcessor) {
```
durbanie, line 40: I think this is a little error prone, and hard to understand.

Instead, I'd split this into two functions that return a fully constructed object, rather than the constructor. At the call-site, you'd just do:

`const processor = buildProcessor_(processorCtorObj, processorCtorParams);`

What are your thoughts?

test/unit/setupMeasure_test.js
```diff
@@ -109,5 +111,28 @@ describe('After calling the setupMeasure function of setup', () => {
             expect(dataLayer.length).toBe(1);
           });
     });
+
+    describe('Does not crash when unsupported strings are passed in', () => {
```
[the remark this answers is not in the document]
Alex7Li: Moved

src/config/configProcessors.js
```diff
@@ -0,0 +1,93 @@
...
+ *     stored, or a string representing an object of this type.
+ * @param {!Object<string, *>} eventOptions Options objects to pass to the
+ *    event processor.
+ * @param {!StorageConstructor|string} storageInterface An interface to a class
+ *     that will handle setting and loading long term data, or a string
+ *     representing an object of this type.
+ * @param {!Object<string, *>} storageOptions Options objects to pass to the
+ *    storage interface.
+ */
+function configProcessors(eventProcessor, eventOptions,
```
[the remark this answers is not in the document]
Alex7Li: Yes

src/config/configProcessors.js
Alex7Li: Sure, that seems easier to understand, especially since it's not likely that the current functions would be called separately.

src/config/configProcessors.js
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * Process an event object by sending it to the registered event processor
+   * along with interfaces to the storage and data layer helper.
+   *
+   * @param {string} name The name of the event to process.
+   * @param {!Object<string, *>=} options The options object describing
+   *     the event.
+   */
+  function processEvent(name, options = undefined) {
```
durbanie, line 121: I think a JSDoc in the middle of a function decreases readability (it looks like a class method) and doesn't provide a whole lot of benefit. I know this function is small, but if you think it needs a JSDoc, please take it out and make it a private module-level function. Otherwise, a normal comment is fine.

Also, for nested functions, please use an arrow function.

Alex7Li: This function uses this and cannot be made an arrow function. I removed the jsdoc.

PR #53, opened 2020-08-05
Add ga event processor class

kjgalvan:
Add the GoogleAnalyticsEventProcessor class with working constructor.

Add tests for constructor to ensure options object parameter destructuring works properly.


karma.conf.js
```diff
@@ -23,6 +23,7 @@ module.exports = function(config) {
       {pattern: 'src/**/*.js'},
       // ------------------------- Test Files ----------------------------------
       'test/*_test.js',
+      'test/**/*_test.js',
```
Alex7Li, line 26: This should just be testing the unit files, why are you including the integration tests too?

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -0,0 +1,129 @@
...
+ */
+class GoogleAnalyticsEventProcessor {
+  /**
+   * @param {{
+   *     'api_secret': (string|undefined),
+   *     'measurement_id': (string|undefined),
+   *     'measurement_url': string,
+   *     'client_id_expires': number,
+   *     'automatic_params': !Object<string, boolean>,
+   * }=} optionsObject
```
Alex7Li, line 19: It seems to me like the first 2 parameters are required, so they can't be undefined. And the last 3 parameters are optional, so they should have an or with undefined. Also, if the first 2 parameters are required, then optionsObject is required and shouldn't be optional. Also I think there should be some documentation about what the parameters mean

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -0,0 +1,129 @@
...
+   * @return {number} How long the key should be stored in seconds.
+   *     If -1, then the default value saved in storage will be used.
+   *     If 0, the data is not saved to long term storage at all.
+   *     If Number.POSITIVE_INFINITY, data will be stored forever.
+   * @export
+   */
+  persistTime(key, value) {
+    // checks to see if key is client ID
+    // Returns client ID expires if true
+    return -1;
```
Alex7Li, line 125: This comment is inaccurate.

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -0,0 +1,129 @@
...
+
+  /**
+   * Processes events pushed to the data layer by constructing and sending JSON
+   * POST requests to Google Analytics. Follows Measurement Protocol.
+   * @param {!StorageInterface} storageInterface An interface to an object to
+   *    load or save persistent data with.
+   * @param {{get:function(string):*, set:function(string, *)}} modelInterface
+   *    An interface to load or save short term page data from the data layer.
+   * @param {string} eventName The name of the event passed to the data layer.
+   * @param {!Object<string, *>=} eventArgs The events passed to the data layer.
```
Alex7Li, line 102: Nit: I wanted to change this name to eventOptions since it is an options object.

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -0,0 +1,129 @@
...
+     * globally in the data model.
+     * @private @const {!Object<string, boolean>}
+     */
+    this.automaticParams_ = Object.assign({
+      'page_path': true,
+      'page_location': true,
+      'page_title': true,
+      'user_id': true,
+      'client_id': true,
+    }, userAutomaticParams);
```
Alex7Li, line 39: This is just a set, right? Eww. (Though I can't think of a better way to do it, maybe leave a comment? Actually thee guidelines say Object.assign adds a polyfill, you can check it https://docs.google.com/document/d/1XIhihp4elYAa9cFqJL9vg3f7u4kbsoZqIQa5ytBqYg0/edit#heading=h.3xoqh97uc7cf

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -0,0 +1,129 @@
...
+   *     'measurement_url': string,
+   *     'client_id_expires': number,
+   *     'automatic_params': !Object<string, boolean>,
+   * }=} optionsObject
+   */
+  constructor({
+      'api_secret': apiSecret,
+      'measurement_id': measurementId,
+      'measurement_url': measurementUrl = 'https://www.google-analytics.com/mp/collect',
+      'client_id_expires': clientIdExpires = Number.POSITIVE_INFINITY,
```
calhounsm, line 44: I feel like there should be some default expiration.  For GA we typically use 2 years.

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -0,0 +1,129 @@
...
+      'user_id': true,
+      'client_id': true,
+    }, userAutomaticParams);
+
+    /**
+     * Parameters that are to be set at the top level of the JSON
+     * POST body instead of as event parameters.
+     * @private @const {!Object<string,boolean>}
+     */
+    this.topLevelParams_ = {
```
calhounsm, line 77: This feels like something that could be a top-level constant rather than a member of the class.

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -0,0 +1,129 @@
+goog.module('measurementLibrary.eventProcessor.GoogleAnalyticsEventProcessor');
+
+/**
+ * A class that processes events pushed to the data layer
+ * by constructing and sending Google Analytics events via
+ * Measurement Protocol.
```
calhounsm, line 6: Would add "Measurement Protocol (App + Web)" to disambiguate that it's not using the version of measurement protocol used by analytics.js.

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -0,0 +1,129 @@
...
+   *     'measurement_id': (string|undefined),
+   *     'measurement_url': string,
+   *     'client_id_expires': number,
+   *     'automatic_params': !Object<string, boolean>,
+   * }=} optionsObject
+   */
+  constructor({
+      'api_secret': apiSecret,
+      'measurement_id': measurementId,
+      'measurement_url': measurementUrl = 'https://www.google-analytics.com/mp/collect',
```
calhounsm, line 43: Could be worth pulling out to a top-level constant, e.g. DEFAULT_MEASUREMENT_URL (especially since it sounds like this will have two possible values depending on compiler flags in a later PR).

src/eventProcessor/GoogleAnalyticsEventProcessor.js
calhounsm: Per conversations with KJ, it may make sense when using a custom measurement_url to not send either the ID, secret, or both.  In this case, they may be added server-side before being forwarded to GA (if they want to forward them at all).

So we decided to make them optional, with a warning that events will be dropped if they are omitted on the default collection path.  I assume the warning will be added in a followup PR.

karma.conf.js
kjgalvan: Sorry, I adjusted it by moving our unit tests into their own folder and adjusting the matcher. I will have a lot of unit tests tied to just this event processor, so for organizational purposes it makes sense to have subfolders for unit testing, like so:

- unit
-- GoogleAnalyticsEventProcessor
--- (bunch of unit tests)
-- uuidv4
--- (a couple unit tests)
- integration

src/eventProcessor/GoogleAnalyticsEventProcessor.js
calhounsm, line 44: This should probably be some finite number.  In GA, we use 2 years as the default expiration, that'd probably work well here as well:
https://developers.google.com/analytics/devguides/collection/gtagjs/cookies-user-id#configure_cookie_field_settings

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -0,0 +1,129 @@
...
+ * @implements {EventProcessor}
+ */
+class GoogleAnalyticsEventProcessor {
+  /**
+   * @param {{
+   *     'api_secret': (string|undefined),
+   *     'measurement_id': (string|undefined),
+   *     'measurement_url': string,
+   *     'client_id_expires': number,
+   *     'automatic_params': !Object<string, boolean>,
```
calhounsm, line 37: It'd be a bit more natural for users if this was an !Array<string>, and the library transforms it internally into a set (for efficiency).  It'd also enforce that the automatic_params list is additive only, which I see as a good thing.

src/eventProcessor/GoogleAnalyticsEventProcessor.js
calhounsm: Per my other comment, if the user params were an array, this could be addressed by changing the code here to:

```
this.automaticParams_ = { /* default auto params */ };
for (let i = 0; i < userAutomaticParams.length; i++) {
  this.automaticParams_[userAutomaticParams[i]] = true;
}
```

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -0,0 +1,129 @@
...
+   * @param {!StorageInterface} storageInterface An interface to an object to
+   *    load or save persistent data with.
+   * @param {{get:function(string):*, set:function(string, *)}} modelInterface
+   *    An interface to load or save short term page data from the data layer.
+   * @param {string} eventName The name of the event passed to the data layer.
+   * @param {!Object<string, *>=} eventArgs The events passed to the data layer.
+   * @export
+   */
+  processEvent(storageInterface, modelInterface, eventName, eventArgs) {
+    // constructs and sends JSON POST requests to GA
```
calhounsm: Nit: Start TODO comments with TODO(kjgalvan):

(Or whoever will do them!)

src/eventProcessor/GoogleAnalyticsEventProcessor.js
calhounsm: +1.  If it's a TODO comment, add TODO(kjgalvan): at the start.

test/GoogleAnalyticsEventProcessor/constructor_test.js
```diff
@@ -0,0 +1,89 @@
...
+
+  it('has the correct default arguments for `measurement_url`, ' +
+  '`client_id_expires`, and `automatic_params`', () => {
+    eventProcessor = new GoogleAnalyticsEventProcessor();
+
+    expect(eventProcessor.measurementUrl_).toBe('https://www.google-analytics.com/mp/collect');
+    expect(eventProcessor.clientIdExpires_).toBePositiveInfinity();
+
+    for (const key in defaultAutomaticParams) {
+      expect(eventProcessor.automaticParams_.hasOwnProperty(key)).toBeTrue();
```
calhounsm: I don't think it's possible for this check to fail, unless something has messed with Object.prototype.

test/GoogleAnalyticsEventProcessor/constructor_test.js
```diff
@@ -0,0 +1,89 @@
...
+
+      if (eventProcessor.automaticParams_.hasOwnProperty(key)) {
+        expect(eventProcessor.automaticParams_[key])
+            .toBe(defaultAutomaticParams[key]);
+      }
+    }
```
calhounsm: I think just expect(automaticParams_).toEqual(defaultAutomaticaParams) is both cleaner and a stronger assertion.

I don't think any of the expectations in this loop would fail if defaultAutomaticParams had a key that was not in the eventProcessor.automaticParams_, but it probably should.

src/eventProcessor/GoogleAnalyticsEventProcessor.js
kjgalvan: So this is a tricky one. In the design doc, [we discussed](https://docs.google.com/document/d/1peiKZs0e7C7n5GS9C1MHzODkHLPGBb55uL8lcqCtXCE/edit#heading=h.2elzgjrj4jsf) why a user would override the measurement URL and if the API secret and measurement ID would be included as query parameters. If, for example, they had a custom endpoint that would take in the events and route them to different Google Analytic properties (each with their own measurement ID) then we definitely don't want to force them to pick a single measurement ID.

The idea we came up with is to keep the API secret and measurement ID as completely optional parameters. If included, they will set the appropriate query parameter during the network request.

Of course for "target" scoping we require a single measurement ID attached to the GA event processor so it can look for the correct object in the global model and extract all its parameters. This will be a documented trade-off; If they decide to forgo a measurement ID then we will trust them to use automatic parameters for persisting important event parameters beyond the defaults.

src/eventProcessor/GoogleAnalyticsEventProcessor.js
kjgalvan: Made this explicitly a TODO

src/eventProcessor/GoogleAnalyticsEventProcessor.js
kjgalvan: Renamed!

test/GoogleAnalyticsEventProcessor/constructor_test.js
```diff
@@ -0,0 +1,89 @@
...
+      'client_id_expires': 0,
+    });
+
+    expect(eventProcessor.clientIdExpires_).toBe(0);
+  });
+
+  it('correctly overrides `automatic_params`', () => {
+    eventProcessor = new GoogleAnalyticsEventProcessor({
+      'automatic_params': {
+        'client_id': false,
```
calhounsm: Not to harp since there are two other comments already, but I feel like letting people do this is just handing them a foot-gun :p.

src/eventProcessor/GoogleAnalyticsEventProcessor.js
kjgalvan: So this is one of the only ways represent a set in JS pre-ES6 without defining a custom data structure. I added more documentation surrounding this parameter that makes it explicitly clear what a user must do to override a default value / add their own.

test/GoogleAnalyticsEventProcessor/constructor_test.js
kjgalvan: I definitely forgot there was an toEqual() matcher 🤦 Updated!

test/GoogleAnalyticsEventProcessor/constructor_test.js
kjgalvan: Fixed! I also realized a bit later when writing some stuff for client ID that it could be very bad to not check globally for it, especially when it gets generated and added to the model globally. Thanks for bringing it up!

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -0,0 +1,155 @@
...
+
+    // Add user provided params to automatic param list
+    for (let i = 0; i < userAutomaticParams.length; ++i) {
+      this.automaticParams_[userAutomaticParams[i]] = true;
+    }
+
+    /**
+     * An API secret that can be generated via Google Analytics UI
+     * NOTE: This processor makes public network requests using this
+     * API secret. To avoid spam, regenerate API secrets regularly.
```
calhounsm, line 81: No action needed in this request, but this note should be made in much more visible documentation for the processor as well!

Alex7Li:
LGTM

PR #54, opened 2020-08-06
Add code to respond to the event function

Alex7Li:
This commit adds code to respond when a user writes `measure('event', 'name', {more: 'params'}). It really just passes this down to the processEvent method of the processor. So +1 line. +engineering

Most of the code in config_test is duplicated from setupMeasure_test. The helper function is somewhat confusing, so it is nicer to be able to access it from the same file in terms of understanding code. Also, the plan is to only use the helper function for setup tests as of now so there isn't a reason to expose it to the other tests. See https://github.com/googleinterns/measurement-library/pull/43#discussion_r465203446


src/setup.js
```diff
@@ -27,9 +27,33 @@ function setupMeasure(dataLayer) {
...
 
-   // TODO: Use the variables declared above
+  helper.registerProcessor('config', configProcessors);
+  helper.process();
+
+
+  // TODO: configProcessors jsdoc and better processor/storage init in PR 52
+  function configProcessors(eventProcessor, eventOptions,
+                            storageInterface, storageOptions) {
+    let processor = new eventProcessor(eventOptions);
```
kjgalvan, line 39: Shouldn't these be constants?

src/setup.js
```diff
@@ -27,9 +27,33 @@ function setupMeasure(dataLayer) {
...
+
+    /**
+     * Process an event object by sending it to the registered event processor
+     * along with interfaces to the storage and data layer helper.
+     *
+     * @param {string} name The name of the event to process.
+     * @param {!Object<string, *>=} args The arguments to
+     *     process with the event.
+     */
+    function processEvent(name, args = undefined) {
```
kjgalvan, line 50: Could you perhaps pass an empty object as the default parameter for args? This will help me not have to type check this argument!

test/integration/setupMeasure_test.js
```diff
@@ -0,0 +1,92 @@
...
+    save = jasmine.createSpy('save');
+    persistTime = jasmine.createSpy('persistTime');
+    processEvent = jasmine.createSpy('processEvent');
+  });
+
+  describe('calls the processEvent function of eventProcessor ' +
+      `with the right arguments after event is called`, () => {
+    executeSnippetBeforeAndAfterSetup(/* config= */(measure) => {
+      measure('config', MockProcessor, {}, MockStorage, {});
+      measure('event', 'name', 'extra arguments');
```
kjgalvan, line 82: NIT: Shouldn't extra arguments be in object with a key value pair?

src/setup.js
Alex7Li: For now, yes

src/setup.js
Alex7Li: The default is an empty object (See line 51). It's written this way because of https://docs.google.com/document/d/1XIhihp4elYAa9cFqJL9vg3f7u4kbsoZqIQa5ytBqYg0/edit#heading=h.x95stkrs2rc2

test/integration/setupMeasure_test.js
Alex7Li: Yep, good catch

src/setup.js
kjgalvan: Today I learned! Thanks Alex :)

src/setup.js
```diff
@@ -27,9 +27,33 @@ function setupMeasure(dataLayer) {
...
+
+    /**
+     * Process an event object by sending it to the registered event processor
+     * along with interfaces to the storage and data layer helper.
+     *
+     * @param {string} name The name of the event to process.
+     * @param {!Object<string, *>=} options The options object describing
+     *     the event.
+     */
+    function processEvent(name, options = undefined) {
```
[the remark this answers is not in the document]
Alex7Li: I think that this goes against the client tagging guidelines https://docs.google.com/document/d/1XIhihp4elYAa9cFqJL9vg3f7u4kbsoZqIQa5ytBqYg0/edit#heading=h.x95stkrs2rc2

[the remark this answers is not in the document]
zikas: Yeah, always felt like we are too extreme there but well, guideline is a guideline.

src/setup.js
```diff
@@ -27,9 +27,33 @@ function setupMeasure(dataLayer) {
...
    * @private @const {!DataLayerHelper}
    */
-  const helper = new DataLayerHelper(dataLayer);
+  const helper = new DataLayerHelper(dataLayer, {'processNow': false});
 
-   // TODO: Use the variables declared above
+  helper.registerProcessor('config', configProcessors);
+  helper.process();
+
+  // TODO: configProcessors jsdoc and better processor/storage init in PR 52
```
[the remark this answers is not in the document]
Alex7Li: I don't know exactly what you mean by reported issue, but I changed the formatting

PR #55, opened 2020-08-06
Create the processSet function

Alex7Li:
This PR adds handling to listen to users that push the command `measure('set', 'key', 'value', ttl=)`. We pass to the storage save function, though there are some extra cases with determining the time to live. By default, we let the eventProcessor decide.


Alex7Li:
I will be several of the unit tests for this pull request once https://github.com/googleinterns/measurement-library/pull/52 is merged, since the new file format will make it easier to test and there is no need to test them all in both orders.

src/config/configProcessors.js
```diff
@@ -109,7 +109,35 @@ function configProcessors(helper, eventProcessor, eventOptions,
...
+ */
+function registerEventAndSet_(helper, processor, eventOptions, storage) {
+  // Check if a given key/value pair should be persisted in storage, and
+  // if so, save it.
+  function processSet(key, value, secondsToLive = undefined) {
+    if (secondsToLive === undefined) {
+      secondsToLive = processor.persistTime(key, value);
+    }
+    // Don't save if the time to live is 0.
+    if (secondsToLive) {
```
durbanie, line 133: Is this only for checking against secondsToLive === 0? What other values could processor.persistTime return?

test/unit/config/set_test.js
```diff
@@ -0,0 +1,85 @@
...
+  it('calls persistTime once with parameters key, value', () => {
+    measure('set', 'a key', 3);
+
+    expect(processor.persistTime).toHaveBeenCalledTimes(1);
+    expect(processor.persistTime).toHaveBeenCalledWith('a key', 3);
+  });
+
+  it('calls save on the storage interface if persistTime returns' +
+      'a positive integer', () => {
+    processor.persistTime.and.returnValue(1337);
```
durbanie, line 33: Noice.

test/unit/config/set_test.js
```diff
@@ -0,0 +1,85 @@
...
+    measure('set', 'key', 'value');
+
+    expect(storage.save).toHaveBeenCalledTimes(1);
+    expect(storage.save).toHaveBeenCalledWith('key', 'value', 1337);
+  });
+
+  it('calls save on the storage interface if persistTime returns ' +
+      'infinity.',
+      () => {
+        processor.persistTime.and.returnValue(
```
durbanie, line 43: Is it possible persistTime could return null/undefined? If so, I'd add a test for it (even just as documentation).

src/config/configProcessors.js
Alex7Li: According to the [contract](https://github.com/googleinterns/measurement-library/blob/master/src/eventProcessor/EventProcessorInterface.js), it is either -1 or a number greater than or equal to 0 (Including infinity). The undefined default value is probably confusing, I'm following [the tagging guideline](https://github.com/googleinterns/measurement-library/blob/master/src/eventProcessor/EventProcessorInterface.js) but I every time I've followed this guideline so far, someone has posted a comment on my PR with some amount of confusion, so it may be a guideline that is better not to follow...

Alex7Li: I added jsdoc to this function because it may help future readers understand the possible values of seconds to live

test/unit/config/set_test.js
Alex7Li: Returning undefined would violate the contract.

src/config/configProcessors.js
Alex7Li: Ah, this recommendation doesn't actually apply since I need an explicit undefined check (the recommendation was to prefer a falsy check), rewriting it actually saves 1 byte.

PR #56, opened 2020-08-06
Add build request url

kjgalvan:
Adds the buildRequestUrl helper function to the GoogleAnalyticsEventProcessor class and relevant unit tests

Also removed unnecessary matcher from karma configuration due to fix console warning.


src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -115,6 +115,27 @@ class GoogleAnalyticsEventProcessor {
     return 'googleAnalytics';
   }
 
+  /**
+   * Builds the POST request URL using `measurement_id` and `api_secret`
+   * query parameters if available
+   * @return {string}
+   * @private
+   */
+  buildRequestUrl() {
```
calhounsm, line 124: I've found that Array.join can simplify URL query string building a lot when some or all of the parameters are optional.

e.g. psuedo-code like:
```
const paramsArray = [];
if (measurementId) paramsArray.push(`measurement_id=${measurementId}`);
// etc.

if (paramsArray.length > 0) {
  const params = paramsArray.join('&');
  // etc..
}
```

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -115,6 +115,27 @@ class GoogleAnalyticsEventProcessor {
     return 'googleAnalytics';
   }
 
+  /**
+   * Builds the POST request URL using `measurement_id` and `api_secret`
+   * query parameters if available
+   * @return {string}
+   * @private
```
calhounsm, line 122: Style guide nit: private methods should end with _.
https://google.github.io/styleguide/jsguide.html#naming-method-names

test/GoogleAnalyticsEventProcessor/buildRequestUrl_test.js
```diff
@@ -0,0 +1,87 @@
...
+    expect(eventProcessor.buildRequestUrl())
+        .toBe(expected);
+  }
+
+  it('creates the correct api hit with default url', () => {
+    checkRequestUrl(
+      /** apiSecret */ '123',
+      /** measurementId */ '456',
+      /** measurementUrl */ undefined,
+      /** expected */ 'https://www.google-analytics.com/mp/collect?measurement_id=456&api_secret=123'
```
calhounsm, line 33: Nit: Line too long.

I usually break URL strings after the "?" first, then after an "&" or "/" if more breaks are required.

test/GoogleAnalyticsEventProcessor/buildRequestUrl_test.js
```diff
@@ -0,0 +1,87 @@
...
+      'measurement_url': measurementUrl,
+    });
+
+    expect(eventProcessor.buildRequestUrl())
+        .toBe(expected);
+  }
+
+  it('creates the correct api hit with default url', () => {
+    checkRequestUrl(
+      /** apiSecret */ '123',
```
calhounsm, line 30: Nit: Param name comments should be in the format `/* name= */ value`. One opening *, and an = after the name.
https://google.github.io/styleguide/jsguide.html#formatting-param-name-comments

test/GoogleAnalyticsEventProcessor/buildRequestUrl_test.js
```diff
@@ -0,0 +1,87 @@
...
+      'api_secret': apiSecret,
+      'measurement_id': measurementId,
+      'measurement_url': measurementUrl,
+    });
+
+    expect(eventProcessor.buildRequestUrl())
+        .toBe(expected);
+  }
+
+  it('creates the correct api hit with default url', () => {
```
calhounsm, line 28: Phrasing is a bit weird here, although tricky since the output is a URL but one of the inputs is too. 

Maybe "creates the correct query string when using the default measurement url" ?

test/GoogleAnalyticsEventProcessor/buildRequestUrl_test.js
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Tests the `buildRequestUrl` function by passing in an API secret,
+   * measurement ID, and measurement URL to construct the event processor
+   * with and comparing the result of the function with the expected result.
+   * @param {string} apiSecret
+   * @param {string} measurementId
+   * @param {string} measurementUrl
+   * @param {string} expected
+   */
+  function checkRequestUrl(apiSecret, measurementId, measurementUrl, expected) {
```
calhounsm, line 17: This is totally optional, but I would probably name this "assertRequestUrl" to highlight that it's a test assertion and more importantly, that it throws.  Or if you really want to fit into the Jasmine syntax, change this to "expectRequestUrl", remove the "expected" param, and return the expectation object.

e.g. calls would look like:
```
expectRequestUrl('secret', 'measurementId', 'measurementUrl').toBe('http://...');
```

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -115,6 +115,27 @@ class GoogleAnalyticsEventProcessor {
...
+   */
+  buildRequestUrl() {
+    const measurementIdQuery =
+        this.measurementId_ ? `measurement_id=${this.measurementId_}` : '';
```
calhounsm, line 126: Since these are arbitrary user inputs, what do you think about passing them through encodeURIComponent?

PR #61, opened 2020-08-12
Add client ID

kjgalvan:
Adds the getClientId_ function to GoogleAnalyticsEventProcessor class with subsequent tests.


src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -115,6 +117,47 @@ class GoogleAnalyticsEventProcessor {
...
+    let clientId = this.getFromGlobalScope_('client_id', modelInterface);
+    if (!clientId || typeof clientId !== 'string') {
+      clientId = storageInterface.load('client_id');
+      if (!clientId || typeof clientId !== 'string') {
+        clientId = uniqueId.generateUniqueId();
+      }
+      storageInterface.save('client_id', clientId, this.clientIdExpires_);
+      modelInterface.set('client_id', clientId);
+    }
+    return clientId;
```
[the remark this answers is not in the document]
kjgalvan: Fixed, thank you!

test/unit/GoogleAnalyticsEventProcessor/getClientId_test.js
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * Builds expectation object for the `getClientId` function by passing
+   * in (if any) the client IDs stored in long term storage and the global
+   * data model.
+   * @param {string|undefined} storedClientId
+   * @param {string|undefined} modelClientId
+   * @return {!Expectation}
+   */
+  function expectClientId(
```
durbanie, line 92: I'm not sure I like this pattern. I think it could be better to reset mockStorage/mockModelInterface in a "beforeEach". Then just do a normal expect on the result of getClientId_.

This is somewhat controversial in the developer community at large, but generally Google prefers test logic to be explicit in most cases. I like the idea of having helper functions for complicated set up or complicated validation (separate), but this helper does setup and executes code under test (as well as calls the expect) in the same function.

For more info, see: 
https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html

src/eventProcessor/generateUniqueId.js
```diff
@@ -0,0 +1,37 @@
+goog.module('measurementLibrary.eventProcessor.generateUniqueId');
+
+/**
+ * Returns a cryptographically secure random number generator if
+ * available.
+ * @return {function(!Uint32Array):!Uint32Array|undefined}
+ */
+function useCryptoRandomValues() {
```
durbanie, line 8: nit: I wouldn't call this "use...", since it suggests a "yes" or "no" result.

test/unit/generateUniqueId/generateUniqueId_test.js
```diff
@@ -0,0 +1,62 @@
...
+    expect(window.crypto.getRandomValues).toHaveBeenCalledTimes(1);
+    expect(Math.random).toHaveBeenCalledTimes(0);
+  });
+
+  it('generates unique ID in correct format using ' +
+      '`Math.random` and `Date.getTime`', () => {
+    spyOn(window.crypto, 'getRandomValues').and.callThrough();
+    spyOn(Math, 'random').and.callThrough();
+
+    const uid = generateUniqueId(null);
```
durbanie, line 33: nit: I'd replace "null" with "undefined", since that's what "useCrypto..." would return.

PR #63, opened 2020-08-13
Add error handling and logging to Event Processor

kjgalvan:
Checks for invalid user input in constructor arguments and sets arguments to default value if detected. Also logs an error when in Debug mode.

Changes default measurement URL to Google Analytics debug server if debug is detected.

Adds tests for these changes.

### DIFF
--- a/src/eventProcessor/GoogleAnalyticsEventProcessor.js
+++ b/src/eventProcessor/GoogleAnalyticsEventProcessor.js
@@ -66,7 +66,7 @@ class GoogleAnalyticsEventProcessor {
       'measurement_id': measurementId,
       'measurement_url': measurementUrl,
       'client_id_expires': clientIdExpires,
-      'automatic_params': userAutomaticParams = [],
+      'automatic_params': userAutomaticParams,
     } = {}) {
     if (apiSecret !== undefined && typeof apiSecret !== 'string') {
       logging.log(
--- a/src/eventProcessor/GoogleAnalyticsEventProcessor.js
+++ b/src/eventProcessor/GoogleAnalyticsEventProcessor.js
@@ -16,6 +16,12 @@ const TOP_LEVEL_PARAMS = {
 };
 
 /**
+ * Default client ID expires value.
+ * @const {number}
+ */
+const DEFAULT_CLIENT_ID_EXPIRES = 2 * 365 * 24 * 60 * 60;
+
+/**
  * Returns the default Google Analytics measurement URL
  * depending on if debug mode has been enabled.
  * When in debug mode, events get sent to the debug Google Analytics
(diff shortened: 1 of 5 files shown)


src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -50,12 +64,57 @@ class GoogleAnalyticsEventProcessor {
...
+    }
+    if (clientIdExpires === undefined || isNaN(Number(clientIdExpires))) {
+      if (clientIdExpires !== undefined) {
+        logging.log(
+          'Provided Client ID Expires is not a valid number. ' +
+              'Using default of two years instead.',
+          logging.LogLevel.ERROR
+        );
+      }
+      clientIdExpires = 2 * 365 * 24 * 60 * 60;
```
zikas, line 110: this should be a named constant

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -50,12 +64,57 @@ class GoogleAnalyticsEventProcessor {
...
+          logging.LogLevel.ERROR
+        );
+      }
+      measurementUrl = getDefaultMeasurementUrl();
+    }
+    if (clientIdExpires === undefined || isNaN(Number(clientIdExpires))) {
+      if (clientIdExpires !== undefined) {
+        logging.log(
+          'Provided Client ID Expires is not a valid number. ' +
+              'Using default of two years instead.',
```
zikas, line 106: I'd just say that we use default value otherwise there's always a risk that we'll change a constant and forget to update the message

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -50,12 +64,57 @@ class GoogleAnalyticsEventProcessor {
...
+      if (measurementUrl !== undefined) {
+        logging.log(
+          'Provided Measurement URL is not a valid string. ' +
+              'Using default URL instead.',
+          logging.LogLevel.ERROR
+        );
+      }
+      measurementUrl = getDefaultMeasurementUrl();
+    }
+    if (clientIdExpires === undefined || isNaN(Number(clientIdExpires))) {
```
zikas, line 102: isNaN(Number(clientIdExpires) should be sufficient

src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -50,12 +64,57 @@ class GoogleAnalyticsEventProcessor {
...
+          'Provided Client ID Expires is not a valid number. ' +
+              'Using default of two years instead.',
+          logging.LogLevel.ERROR
+        );
+      }
+      clientIdExpires = 2 * 365 * 24 * 60 * 60;
+    } else {
+      clientIdExpires = Number(clientIdExpires);
+    }
+    if (Object.prototype.toString.call(userAutomaticParams) !==
```
zikas, line 114: While this is correct, could you please check what browser versions we must support. There's a chance that Array.isArray will do

src/eventProcessor/GoogleAnalyticsEventProcessor.js
kjgalvan: I also am using these blocks to also set default values - I need to detect is client_id_expires was undefined as Number(undefined) actually becomes a number (0).

PR #64, opened 2020-08-13
Merge scopes together when calling the event command

Alex7Li:
This commit updates the event command. The final parameter should include not just those parameters defined in the command, but also those defined by calling the command `measure('set, 'processorName', {paramsTo: 'persist'})`, and those defined in the constructor.

There are some issues with the set command that cause it not to work for now, explained here: https://docs.google.com/document/d/1BRUjcRSOG2cX1okcjbqsXlCM62wozotrvniEYjz736Q/edit#
It should work when data layer helper is updated (https://github.com/google/data-layer-helper/pull/59)

### DIFF
--- a/src/config/setup.js
+++ b/src/config/setup.js
@@ -30,9 +30,10 @@ function setupMeasure(dataLayer) {
   const helper = new DataLayerHelper(dataLayer, {'processNow': false});
 
   helper.registerProcessor('config',
-      (eventProcessor, eventOptions, storageProcessor, storageOptions) =>
+      (eventProcessor, eventOptions, storageProcessor, storageOptions) => {
         configProcessors(helper, eventProcessor, eventOptions,
-            storageProcessor, storageOptions));
+            storageProcessor, storageOptions);
+      });
   helper.process();
 }
 
--- a/src/config/setup.js
+++ b/src/config/setup.js
@@ -30,10 +30,9 @@ function setupMeasure(dataLayer) {
   const helper = new DataLayerHelper(dataLayer, {'processNow': false});
 
   helper.registerProcessor('config',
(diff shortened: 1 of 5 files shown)


src/config/setup.js
```diff
@@ -30,10 +30,9 @@ function setupMeasure(dataLayer) {
   const helper = new DataLayerHelper(dataLayer, {'processNow': false});
 
   helper.registerProcessor('config',
-      (eventProcessor, eventOptions, storageProcessor, storageOptions) => {
+      (eventProcessor, eventOptions, storageProcessor, storageOptions) =>
```
durbanie, line 33: Why did you remove the braces?

test/unit/config/processEvent_test.js
```diff
@@ -0,0 +1,107 @@
...
+    dataLayer = [];
+    measure = function() {
+      dataLayer.push(arguments);
+    };
+    setupMeasure(dataLayer);
+  });
+
+
+  describe('The call forwarded after a call to measure(`event`, ...)', () => {
+    /* TODO:(wolfblue@): Fix the issue here, it stems from data layer helper
```
durbanie, line 48: I would strongly prefer that we don't merge commented code, so please uncomment once https://github.com/google/data-layer-helper/pull/59 is merged and DLH is updated.

src/config/setup.js
Alex7Li: There wasn't really any reason, I put them back in so we don't have random changes in the commit history

test/unit/config/processEvent_test.js
Alex7Li: It has been uncommented! I will save WIP code locally from now on

PR #65, opened 2020-08-13
Add process event

kjgalvan:


src/eventProcessor/GoogleAnalyticsEventProcessor.js
```diff
@@ -9,11 +9,11 @@ const logging = goog.require('measurementLibrary.logging');
  * @const {!Object<string,boolean>}
  */
 const TOP_LEVEL_PARAMS = {
-  'client_id': true,
-  'user_id': true,
-  'timestamp_micros': true,
-  'user_properties': true,
-  'non_personalized_ads': true,
+  'clientId': true,
```
calhounsm, line 12: I assume that the motivation for changing to camel case is because the JSON body uses camel case for these values?

I'd actually rather stick to snake-case for inputs to maintain cross-compatibility with gtag's schema in cases where the keys are the same (e.g. `user_id` and `user_properties` are documented keys in gtag's schema).  Maybe adjusting this to a string->(string|boolean) map for this and automatic params, where a string is the alternate key for transmission.  No need to adjust the config option for additional params though.

test/unit/GoogleAnalyticsEventProcessor/processEvent_test.js
```diff
@@ -0,0 +1,319 @@
...
+        jasmine.Ajax.requests.mostRecent().data().events[0].params.page_path
+      ).toBe('model_page_path');
+    });
+
+    it('has event level global values when key is defined in model and ' +
+        'is added to automatic parameter list', () => {
+      eventProcessor.processEvent(
+        emptyStorage,
+        new MockModelInterface({
+          added_param: 'model_added_param',
```
calhounsm, line 173: I would add a "not_added_param" to the global model, either here or in a separate test, and an assert that it doesn't get added.

test/unit/GoogleAnalyticsEventProcessor/processEvent_test.js
```diff
@@ -0,0 +1,319 @@
...
+        'that are defined', () => {
+      const eventOptions = {
+        param_one: 1,
+        param_two: '2',
+        param_three: null,
+        param_four: {
+          param_five: true,
+        },
+        param_six: ['test'],
+        param_seven: false,
```
calhounsm, line 140: I'd add a parameter that is explicitly undefined and assert that it's not sent.

PR #68, opened 2020-08-17
Add measure to demo

Alex7Li:
This PR is mostly the demo folder of KJ's branch of measurement library that we used for the demo, found here https://github.com/kjgalvan/measurement-library/tree/complete-measure-lib 

However, her version only sent events to measure, I modified it to send events to gtag as well and cleaned up the code a bit.

A couple steps happened: 
1) When kj added measure, there were some errors with involving the build. The workaround was to explicitly write every file used separately in the gruntfile, rather than using pattern matching. It looks ugly, but it works.
2) Adding a measure file to the public folder of demo. However, the normal measure file would not work, since both measure and gtag are built on the data layer, and gtag reacts to commands like 'event'. So we need them to be listening to a different data layer. Hence, the measure file was generated without calling the main.js file that automatically grabs the window.dataLayer. To do this, another configuration in the gruntfile was needed, and another main file was needed. Initially I tried to compile from the src/config/setup file, but the closure compiler had issues with that approach.
3)  KJ added code to send events from within the library to measure and changed the code snippets to reflect this.
4)  I add calls to gtag back in


demo/public/index.html
```diff
@@ -1,11 +1,27 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <!-- Include gtag... data layer commands conflict with measure though -->
```
calhounsm, line 4: I think this is fine.  If anything, it shows that measurement library can send approximately the same events to GA as gtag would.

demo/public/index.html
```diff
@@ -1,11 +1,27 @@
...
+         conflicting events. As a workaround, we do not run async and include
+         a measure-setup compiled file that does not run commands on load,
+         then call it in the setup script. -->
+    <script src="./measure-no-config.js"></script>
+    <script>
+      measureDataLayer = [];
+      window.setupMeasure(measureDataLayer);
+      function measure(){measureDataLayer.push(arguments);}
+      measure("config", "googleAnalytics", {
+            "measurement_id": "%REACT_APP_GA_TRACKING_ID%",
```
calhounsm, line 22: I actually would prefer using a separate ID, so that we can more easily compare which events are from ML and which are from standard gtag.  I'll set up a property for this.

demo/src/utils.js
```diff
@@ -15,23 +15,16 @@ export function deepCopy(data) {
...
  * @param {string} event
  * @param {!Object<string, *>} parameters
+ * @param {string=} library
  * @return {string} Code snippet of the event.
  */
-export const getEventCodeSnippet = (event, parameters) => {
-  return `gtag("event", "${event}", ` +
+export function getEventCodeSnippet(event, parameters, library = "gtag") {
+  return `${library}("event", "${event}", ` +
       `${JSON.stringify(parameters, undefined, 2)})`;
```
calhounsm, line 26: It looks above like the measure() snippets should include an additional page_title param that isn't in gtag (because it's automatically added).

demo/public/index.html
Alex7Li: Hmm, I think that you are right and it might work with the same data layer. Though I already made them separate and I think separate is a bit better, so I won't change it back. I think separate is better because 1) it doesn't log errors to the console like below and 2) we don't send duplicate events. Technically we could just send 1 message to the data layer for each event instead of 1 per library but I feel like that could be limiting if measurement library changes to become less like gtag in the future

image

demo/public/index.html
Alex7Li: Oh, this is a great idea actually. I don't see any property right now but once it exists I will update the PR

demo/.env
```diff
@@ -1 +1,4 @@
-REACT_APP_GA_TRACKING_ID=G-70CMQK5VQE
\ No newline at end of file
+REACT_APP_GTAG_TRACKING_ID=G-70CMQK5VQE
+REACT_APP_GTAG_API_SECRET=e45cwxLtQ46v5dLqr9v5dw
```
calhounsm, line 2: gtag doesn't use the API secret for transmission (it uses a different wire protocol that doesn't require one), so this could be removed from the environment.  Probably not much harm in leaving it, but just thinking on the principle of not putting more into this then we need to.

demo/public/index.html
calhounsm: Yeah, I'm fine leaving them in a separate data layer mostly due to the different config command formats (although it'd also be cool to see them process the same data layer into similar events).  Was just noting that the comment was a bit misleading :).

demo/.env
Alex7Li: Ah, I didn't notice that it didn't need the api secret, that's interesting. I removed the line.

PR #71, opened 2020-08-17
Cookie save

pedrosilva0111:
Implements save method for CookiesStorage


src/storage/CookiesStorage.js
```diff
@@ -67,18 +67,46 @@ class CookiesStorage {
...
       `registrable_domain=;domain=${domain};expires=${pastDate}`;
 
     return domain;
   }
 
   /** @override */
-  save(key, value, secondsToLive) {}
+  save(key, value, secondsToLive =
+    /** @type {?number} */ (this.settings_['expires'])) {
+    const valString = JSON.stringify(value);
```
Alex7Li: Should this cast be to `number` (no ?)?

src/storage/CookiesStorage.js
```diff
@@ -47,7 +47,7 @@ class CookiesStorage {
   /**
    * Finds the registrable domain of current URL and sets it as the default
    *    domain for cookies
-   * @param {!document} document the document whose cookies are to be set.
+   * @param {!Object} document the document whose cookies are to be set.
    * @return {?string} the registrable domain of docDomain
    * @private
```
Alex7Li, line 52: Nit: The 3 sentences in this JSdoc should start with a capital letter and end in a period. Also check the rest of this file

src/storage/CookiesStorage.js
```diff
@@ -67,18 +67,46 @@ class CookiesStorage {
...
+    newCookie = this.addSettings(newCookie, secondsToLive);
+    document.cookie = newCookie;
+  }
+
+  /**
+   * Finds the registrable domain of current URL and sets it as the default
+   *    domain for cookies
+   * @param {string} cookie the cookie to add settings to.
+   * @param {?number} secondsToLive number of seconds cookie will be in storage.
+   * @return {string} the cookie with all added settings.
```
Alex7Li, line 94: Add period/caps to this jsdoc too

src/storage/CookiesStorage.js
```diff
@@ -67,18 +67,46 @@ class CookiesStorage {
...
+   * @private
+   */
+  addSettings(cookie, secondsToLive) {
+    cookie = `${cookie}; domain=${this.settings_['domain']}`;
+
+    const expiry = new Date();
+    expiry.setMilliseconds(expiry.getMilliseconds() + (secondsToLive + 1) *
+      1000);
```
Alex7Li, line 102: Why are you adding 1 here? It's [documented](https://github.com/googleinterns/measurement-library/tree/5b17e8fbe7341eadd67fc16964f8b2f0ffb5e5a1#set-command) that if secondsToLive is 0, then the value should not be saved, if that is your concern. I thought it was written [here](https://github.com/googleinterns/measurement-library/blob/5b17e8fbe7341eadd67fc16964f8b2f0ffb5e5a1/src/storage/StorageInterface.js) but apparently not, might be worth adding that in.

Alex7Li: Actually, NUMBER.POSITIVE_INFINITY may present an edge case for you, since cookies can't be set to never expire. Maybe add a check and store it for several years?

Alex7Li:
LGTM

test/unit/CookiesStorageTests/cookieSave_test.js
```diff
@@ -0,0 +1,52 @@
+goog.module('measurementLibrary.storage.CookiesStorage.testing.cookieSave');
```
calhounsm, line 1: I know the PR description notes that it's not possible to directly validate cookie settings (expiry, domain) were set correctly, and that's true (at least without significant additional tooling like Selenium tests).

That said, I have a suggestion for a medium sized tweak that would allow at least some weaker tests to be written.  However, it's a decent change and I'm fine if you'd like to merge this PR without it and consider it for a followup PR.

There are a few tricks that can be used to at least record the full "write" string (so, including expiry, domain, and other settings) for indirect verification.  While there's still no way you can verify that settings are interpreted correctly by the browser, you can at least verify that your code generates set strings that *should* be generated correctly on *most* browsers.  Since the cookie API is so stable, this is still a pretty good outcome.

On to the tricks.  My favorite approach is to have a dead-simple "setCookie(cookieString)" method that can be replaced with a spy for unit tests.  So e.g.:

```
save(...) {
  // Build cookie string as in PR.
  setCookie(cookie);
}

/** @param {string} cookieStr */
setCookie(cookieStr) {
  document.cookie = cookieStr;
}
```

Then in unit tests, setCookie can be replaced with a spy and you can assert that it was called with an expected string.

pedrosilva0111: I like this idea and I think it's important to include in this PR. I looked into it and I think it would be nice use this:
spyOn(storage, "setCookie").and.callThrough();
storage.save(...);
const correctCookieString = ...;
expect(storage.setCookie).toHaveBeenCalledWith(correctCookieString);

Is this the kind of spy you think would work?

test/unit/CookiesStorageTests/cookieSave_test.js
```diff
@@ -0,0 +1,52 @@
...
+goog.setTestOnly();
+
+const CookiesStorage = goog.require('measurementLibrary.storage.CookiesStorage');
+
+describe('The save method for cookiesStorage', () => {
```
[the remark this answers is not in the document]
pedrosilva0111: So the tests that are included now have the default prefix (ml_) I will add some tests for empty prefix and one more non-empty custom prefix.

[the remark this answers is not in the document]
calhounsm: Right, I just meant that since we provide an option to override that, there should be at least one test with a custom prefix.  And testing +1 testing the behavior when it's set to empty is a good idea too!

test/unit/CookiesStorageTests/cookieSave_test.js
calhounsm: Yeah, that's pretty much exactly what I had in mind.  The fact that Jasmine's spys make it easy to call through so you can verify at least the part of the actual behavior that you can with the cookie API is a nice perk, though!

test/unit/CookiesStorageTests/cookieSave_test.js
```diff
@@ -33,20 +64,125 @@ describe('The save method for cookiesStorage', () => {
...
+    });
+
+    fullAttributeStorage.save('hello', 'goodbye');
+
+    expect(testCookie.includes(`prehello="goodbye"`)).toBe(true);
+
+    expect(testCookie.includes(`domain=${document.domain}`)).toBe(true);
+  });
+
+  it('Saves cookies with correct expiry', () => {
```
calhounsm, line 155: tl;dr: I'd recommend faking out date using Jasmine's mock clock feature:
https://jasmine.github.io/api/edge/Clock.html

Long version with rationale and stuff:
This test is pretty good.  It verifies that the expiry is approximately correct (set to within 1 second of the user config), and should pass the vast majority of times it's run.  It may fail on a particularly slow machine, where the time difference between when maxExpected is recorded and when save does a similar computation is > 1 second, but runs like that will be very rare.

IMO, #1 is fine.  +/-1 second is pretty small relative to the time durations we expect to see for expirations (usually on the order of years, and probably not less than ~24 hours).

#2 is a small problem for a single test (you can rerun it a couple times to confirm that it was just a one-time flake), but gets to be a bigger problem in larger projects with many tests that have similar issues.

That said, it'd be a stronger test if it:
- Faked out the Date constructor while the code-under-test is executed.
- Had the fake date constructed always at a fixed time.
- Assert that the cookie string's expiry section is always the fixed time + the user-configured expires time.

Date is a pain to fake out by hand though.  But there's good news!  Jasmine has a handy-dandy utility for it.  They have some documentation at the link in the tl;dr, and it seems pretty easy to use.

pedrosilva0111: Thank you for the advice!

PR #72, opened 2020-08-18
Cookie load

pedrosilva0111:
Adds load method for cookie storage.
Note: Once both save and load are merged, additional tests will be added to demonstrate the two working together.


src/storage/CookiesStorage.js
```diff
@@ -67,21 +67,67 @@ class CookiesStorage {
...
   /** @override */
-  load(key, defaultValue) {}
+  load(key, defaultValue) {
+    if (!document.cookie.includes(`${key}=`)) {
+      return defaultValue;
+    }
+
+    const cookieValueJSON = document.cookie.split('; ')
+      .find((row) => row.startsWith(`${this.settings_['prefix']}${key}`))
+      .split('=')[1];
```
Alex7Li, line 127: This line is a bit complicated, I would either split it up or add a comment above.

test/unit/CookiesStorageTests/cookieLoad_test.js
```diff
@@ -0,0 +1,39 @@
...
+goog.setTestOnly();
+
+const CookiesStorage = goog.require('measurementLibrary.storage.CookiesStorage');
+
+describe('The load method for cookiesStorage', () => {
+  const storage = new CookiesStorage({});
+
+  const addCookie = (key, value) => {
+    const expiry = new Date();
+    expiry.setMilliseconds(expiry.getMilliseconds() + 1000);
```
Alex7Li, line 11: Setting the expiration time and hoping that the code finishes in 1 second is a little sus. I think it works fine here, but perhaps add a comment explaining it since it's a bit weird?

src/storage/CookiesStorage.js
```diff
@@ -67,21 +67,67 @@ class CookiesStorage {
...
+    const pastDate = 'Thu, 01 Jan 1970 00:00:00 GMT';
+    document.cookie =
       `registrable_domain=;domain=${domain};expires=${pastDate}`;
 
     return domain;
   }
 
   /** @override */
-  save(key, value, secondsToLive) {}
+  save(key, value, secondsToLive =
```
zikas, line 81: Should this method have a jsdoc? I don't remember the guidance on overrides, please double-check that no jsdoc is necessary (I guess I'm not following why API now redeclares secondsToLive as  parameter).

src/storage/CookiesStorage.js
```diff
@@ -67,21 +67,67 @@ class CookiesStorage {
...
+    /** @type {number} */ (this.settings_['expires'])) {
```
zikas, line 82: Is it intentional that the guideline isn't followed here?
https://docs.google.com/document/d/1XIhihp4elYAa9cFqJL9vg3f7u4kbsoZqIQa5ytBqYg0/edit#heading=h.x95stkrs2rc2

zikas, line 82: Is this a user-provided value?, why are we sure that this is a valid number/not undefined?

src/storage/CookiesStorage.js
```diff
@@ -67,21 +67,67 @@ class CookiesStorage {
...
+   * @return {string} The cookie with all added settings.
+   * @private
+   */
+  addSettings(cookie, secondsToLive) {
+    cookie = `${cookie}; domain=${this.settings_['domain']}`;
+
+    const expiry = new Date();
+
+    if (secondsToLive === Number.POSITIVE_INFINITY) {
+      const tenYears = 10 * 365 * 24 * 60 * 60;
```
zikas, line 104: Please use constant for this value

src/storage/CookiesStorage.js
```diff
@@ -67,21 +67,67 @@ class CookiesStorage {
...
+      expiry.setMilliseconds(expiry.getMilliseconds() + (tenYears) *
+      1000);
+    } else {
+      expiry.setMilliseconds(expiry.getMilliseconds() + (secondsToLive + 1) *
```
zikas, line 108: Why +1 ? it's confusing and if necessary, definitely requires documentation

src/storage/CookiesStorage.js
```diff
@@ -67,21 +67,67 @@ class CookiesStorage {
...
+      1000);
+    }
+
+    cookie = `${cookie}; expires=${expiry.toUTCString()}`;
```
zikas, line 112: would += be more readable?
The alternative is to combine all string building into as single instruction,
cookie = `${cookie}; domain=...; expires=...; ${this.settings}`

src/storage/CookiesStorage.js
```diff
@@ -67,21 +67,67 @@ class CookiesStorage {
...
+   * @private
+   */
+  addSettings(cookie, secondsToLive) {
+    cookie = `${cookie}; domain=${this.settings_['domain']}`;
+
+    const expiry = new Date();
+
+    if (secondsToLive === Number.POSITIVE_INFINITY) {
+      const tenYears = 10 * 365 * 24 * 60 * 60;
+      expiry.setMilliseconds(expiry.getMilliseconds() + (tenYears) *
```
zikas, line 105: Why .setMillisecons in both if branches? Maybe try to just calculate the duration like 
```
secondsToLive === ... ? tenYears : secondsToLive
```
and then
```
expiry.setMilliseconds(... + secondsToLive * 1000)
```
can be called ones

src/storage/CookiesStorage.js
```diff
@@ -67,21 +67,67 @@ class CookiesStorage {
...
+
+    cookie = `${cookie}; ${this.settings_['flags']}`;
+
+    return cookie;
+  }
 
   /** @override */
-  load(key, defaultValue) {}
+  load(key, defaultValue) {
+    if (!document.cookie.includes(`${key}=`)) {
```
zikas, line 121: Includes isn't supported in IE
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

so either it will fail or we'll incur a polyfill

src/storage/CookiesStorage.js
```diff
@@ -67,21 +67,67 @@ class CookiesStorage {
...
+      return defaultValue;
+    }
+
+    const cookieValueJSON = document.cookie.split('; ')
```
zikas, line 125: Did you mean to have '; ' instead of ';' ? I'd prefer trimming value instead otherwise it will highly depend on the string being formatted properly which is usually not a requirement for regular cookie strings

src/storage/CookiesStorage.js
```diff
@@ -67,21 +67,67 @@ class CookiesStorage {
...
+      .find((row) => row.startsWith(`${this.settings_['prefix']}${key}`))
```
zikas, line 126: Startswith is not supported in IE
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
please use indexOf instead

src/storage/CookiesStorage.js
zikas: +1 on this, too much logic in a single instruction

PR #77, opened 2020-08-24
Cookie Load

pedrosilva0111:


src/storage/CookiesStorage.js
```diff
@@ -127,7 +127,17 @@ class CookiesStorage {
...
 
   /** @override */
-  load(key, defaultValue) {}
+  load(key, defaultValue) {
+    if (!document.cookie.includes(`${key}=`)) {
+      return defaultValue;
+    }
+
+    const cookieValueJSON = document.cookie.split('; ')
+      .find((row) => row.startsWith(`${this.settings_['prefix']}${key}`))
```
calhounsm, line 136: Per cookie syntax, each cookie (`row` here) *should*, although may not always, start with a space, and may contain additional whitespace around the cookie name.

https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#Syntax

So, these are all valid and equivalent cookies that the code should handle:

```
"my_cookie=my_value"
" my_cookie=my_value"
"     my_cookie    = my_value "
```

src/storage/CookiesStorage.js
```diff
@@ -127,7 +127,17 @@ class CookiesStorage {
...
+      .split('=')[1];
```
calhounsm, line 137: Since the value of the cookie might contain an "=" as well, you'll want to return all the remaining values in that case.

Example cookie:
```
"my_base64_value=bGV0J3Mgc2VlIGlmIGFueW9uZSBkZWNvZGVzIHRoaXM="
```

src/storage/CookiesStorage.js
pedrosilva0111: Instead of startswith, I changed it to indexOf(prefix + key) >= 0;

This recognizes the cookie even with whitespace, and JSON is not affected by it.
Does that seem ok?

calhounsm: I think this will run into trouble if a cookie's name appears in the value of another cookie, or if two cookies have names where one is a subset of the other.

A few more example cookie strings:
```
"my_cookie=my_value; my_cookie_2=my_other_value"
"my_cookie=my_value; my_favorite_cookie=my_cookie"
// And the browser may have things ordered in reverse.
```

In either case, if I ask for the value of "my_cookie", I might get something unexpected back.

